PR #1, opened 2025-06-21
Prettify: JSON content prettified in prices.json

fswair:
I changed prices.json for better understanding.

And first (also least) PR welcome!


samuelcolvin:
Thanks, but this file is designed to be as small as possible, if you want to read the prices, looking in `providers/`.

PR #22, opened 2025-07-08
Python package

samuelcolvin:


samuelcolvin:
* source abc
* remove`async_` prefix
* use protocol for `Usage`
* call it `auto_update_` not `phone_home_`

PR #37, opened 2025-07-13
checking prices with zenrows and Pydantic AI

samuelcolvin:
Thanks to a suggestion from @fswair, to use zenrows.com to get HTML with js already run, we're able to prices which can be processed by pydantic-ai.

This is a WIP, still needs:
* [ ] more testing
* [ ] check tiered prices are working
* [ ] update yaml in place
* [ ] add a tag saying when the prices were checked by AI
* [ ] ultimately we want to run this as a github action weekly


fswair:
It's nice to being helpful about this! Appreciate your work, always here to help about !

PR #77, opened 2025-08-01
Separate input output

GSemikozov:
@GSemikozov please replicate this to typescript, in this PR.

Also remove requests from usage - it makes no sense.

### DIFF
--- a/packages/js/src/types.ts
+++ b/packages/js/src/types.ts
@@ -6,7 +6,6 @@ export interface Usage {
   input_audio_tokens?: number
   cache_audio_read_tokens?: number
   output_audio_tokens?: number
-  requests?: number
 }
 
 export interface Tier {
--- a/packages/js/src/__tests__/calcPrice.test.ts
+++ b/packages/js/src/__tests__/calcPrice.test.ts
@@ -69,20 +69,19 @@ describe('Core Price Calculation Function', () => {
       const usage: Usage = {
         input_tokens: 1000,
         output_tokens: 500,
-        requests: 2,
       }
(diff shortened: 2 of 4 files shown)


petyosi:
Pulling my comments from the previous review:

- Let's use camelCase; 
- use `matchObject` in vitest to reduce the assertions;
- Use `test.for/it.for` to create repeated table tests;

GSemikozov:
> Pulling my comments from the previous review:
> 
> * Let's use camelCase;
> * use `matchObject` in vitest to reduce the assertions;
> * Use `test.for/it.for` to create repeated table tests;

@samuelcolvin in one of previous PR's we agreed to remove `snake_case -> camelCase` converter for js and simply use the same field names as in json. WDYT about @petyosi suggestion regarding camelCase in ts types?

packages/js/src/types.ts
```diff
@@ -77,8 +76,16 @@ export interface Provider {
...
+  input_price: number
+  output_price: number
+  total_price: number
+}
+
 export interface PriceCalculation {
-  price: number
+  input_price: number
+  output_price: number
+  total_price: number
```
samuelcolvin, line 88: I'm fine to use camel case here, but we obviously can't change the interface to the providers data.

GSemikozov: fyi @petyosi

petyosi: That makes sense.

GSemikozov: upd: agreed not to make this update to have a good consistency with python and data.json.

PR #80, opened 2025-08-04
WIP: common release management

GSemikozov:
common release management for all packages (currently - python + js)


README.md
```diff
@@ -86,13 +86,37 @@ project wherever you use it and [contribute](#contributing) back to the project
 
 ### Release Workflow
```
samuelcolvin, line 87: this should go in a separate file `release.md`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,128 @@
...
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
samuelcolvin, line 21: we really don't need `name:` for stuff like checkout.

samuelcolvin (changes requested):
I don't understand what's going on here, but I think you're massively over complicating things, and it has big drawbacks like there's no change log generated by github

Either:
* we manually update the version in `pyproject.toml` and `package.json`, push, then create a release - I've done this for years, it takes minutes and works fine
* or, update the versions when a release is created, if we're doing that we should follow the same process we use in pydantic-ai rather than reinventing the wheel

PR #129, opened 2025-08-26
Google vertex usage

samuelcolvin:
See [this internal slack thread](https://pydantic.slack.com/archives/C0962TG6YLT/p1756137129352419) for some context.


packages/js/src/types.ts
```diff
@@ -54,6 +54,13 @@ export type MatchLogic =
   | { regex: string }
   | { starts_with: string }
 
+export interface FindItem {
```
petyosi, line 57: if I understand the usage of this correctly, this looks more like
```ts
interface ArrayMatch {
  field: string
  itemMatch: MatchLogic
}
```

petyosi (approved):
We can merge this. I feel like a rename of the FindItem would make its purpose clearer, but I'm happy to do that in a separate PR.

packages/js/src/types.ts
samuelcolvin: done.

PR #131, opened 2025-08-26
Upgrade uv_build to latest to fix wheel permissions issue

edaniels:
See https://github.com/astral-sh/uv/issues/14920

### DIFF
--- a/packages/python/pyproject.toml
+++ b/packages/python/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["uv_build>=0.7.13,<0.8"]
+requires = ["uv_build>=0.8.10"]
 build-backend = "uv_build"
 
 [project]


edaniels:
@samuelcolvin can you take a look at this? Thanks!

samuelcolvin:
Thanks so much, do you need a release?

edaniels:
A release would be great but I'm not sure if that means pydantic ai will pick it up (haven't checked the constraints)

edaniels:
@samuelcolvin ideally, you could release this and then pydantic_ai_slim should pick it up. Side note, For whatever reason, using the dynamic versioning for this dep (https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pyproject.toml#L55-L66) causes uv.lock to not pick up the dependency.

Issue #146, opened 2025-09-17
Is the `pydantic>=2.11.7` dependency requirement necessary?


jerry-heygen:
Currently `genai-prices` depends on `pydantic>=2.11.7`. This has effectively made `pydantic-ai-slim` depend on 2.11.7 despite that it only explicitly requires `pydantic>=2.10` [here](https://github.com/pydantic/pydantic-ai/blob/66fa21ab88b92cddba2734ef68c06c0d510bf92a/pydantic_ai_slim/pyproject.toml#L59).

Is there a reason that `genai-prices` needs 2.11.7? If so, we should update the dependency requirement on `pydantic-ai-slim`; if not, we should probably loosen the version requirement here.

jerry-heygen:
@DouweM #147 Please take a look. Thank you!

PR #147, opened 2025-09-19
Relax pydantic dependency requirement to 2.10

jerry-heygen:
As discussed in #146, this PR lowers the version requirement for `pydantic` in the dependency from 2.11.7 to 2.10.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -311,7 +311,7 @@ dependencies = [
 requires-dist = [
     { name = "eval-type-backport", marker = "python_full_version < '3.11'", specifier = ">=0.2" },
     { name = "httpx", specifier = ">=0.28.1" },
-    { name = "pydantic", specifier = ">=2.11.7" },
+    { name = "pydantic", specifier = ">=2.10" },
 ]
 
 [[package]]
--- a/packages/python/pyproject.toml
+++ b/packages/python/pyproject.toml
@@ -35,7 +35,7 @@ classifiers = [
 dependencies = [
     "eval-type-backport>=0.2; python_version < '3.11'",
     "httpx>=0.28.1",
-    "pydantic>=2.11.7",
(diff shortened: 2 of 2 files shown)


DouweM:
@jerry-heygen Can you regenerate `uv.lock` with 2.10, so we use it in CI? Right now the CI jobs still show `pydantic==2.11.7`.

jerry-heygen:
Sure, I've change the `prices/pyproject.toml` to point to `pydantic>=2.10` as well.

uv.lock
```diff
@@ -779,7 +779,7 @@ dependencies = [
 [package.metadata]
 requires-dist = [
     { name = "httpx", specifier = ">=0.28.1" },
-    { name = "pydantic", specifier = ">=2.11.7" },
```
DouweM, line 782: @jerry-heygen Further down, it still says:

```
name = "pydantic"
version = "2.11.7"
```

Are you editing the file manually? :) Can you please use `uv add pydantic==2.10.0 && uv lock` to regenerate it? (I think that's the right command, but it may need tweaking)

jerry-heygen: No I was using `uv add "pydantic>=2.10" && uv sync` - but sure I can `pip install` the 2.10 version and lock it.

DouweM: @jerry-heygen Ah yeah, I want to make sure we lock the oldest version so it's used in CI (alternatively we could add a CI step that explicitly installs the lowest versions supported, but that's more work than I want to ask you to do in this PR!)

DouweM:
@jerry-heygen Can you run `make pre-commit` to see what files it wants to update, and push those up in the PR?

samuelcolvin (approved):
LGTM! thanks so much.

@DouweM do you want me to merge this and  do a release?

PR #159, opened 2025-10-14
enh: add openai embedding usage extractor

tzing:
Add extractor for the usage model from OpenAI's [/v1/embeddings](https://platform.openai.com/docs/api-reference/embeddings/create) response


alexmojaki:
Thanks!

PR #183, opened 2025-11-06
Remove 'packages/python' workspace from package.json to allow running deno

alexmojaki:
Some `make` commands use `deno`, which would complain:

```
error: Could not find package.json for workspace member in 'file:///Users/alex/work/genai-prices/packages/python/'.
```

which is fair, I don't know why it was a workspace.


petyosi (approved):
I don't think using deno for the helicone sync is necessary. However, the line you're removing is also a bogus leftover, so 👍🏼 .

PR #226, opened 2025-11-25
chore(*): soften httpx python dependency

forforeach:


forforeach:
Oh, sorry, it is not locked there, so basically it might be used there immediately! Thanks

PR #229, opened 2025-11-27
Fix context windows for OpenAI embedding models

DouweM:
See https://platform.openai.com/docs/guides/embeddings#embedding-models

The original numbers were the length of the returned vector, not the context window (max input size)


DouweM:
@alexmojaki Yep I did this from the GitHub web interface so knew I still had a few things to do :) I also plan to update this PR with prices for Cohere embedding models (https://docs.cohere.com/docs/cohere-embed), unless you prefer separate PRs

PR #240, opened 2025-12-10
Add Cohere Embed v4.0

DouweM:
Cohere has more embeddings models (https://docs.cohere.com/docs/cohere-embed), but no prices on https://cohere.com/pricing...

I'm not sure if adding `api_flavor`s requires any changes in Pydantic AI or Gateway.


alexmojaki:
Thanks!

PR #330, opened 2026-03-30
Add cache_write_tokens mapping to UsageExtractor

ali-corpo:
Add cache_write_tokens mapping to UsageExtractor

### DIFF
--- a/packages/python/genai_prices/data.py
+++ b/packages/python/genai_prices/data.py
@@ -929,6 +929,9 @@ providers: list[Provider] = [
                         path=['prompt_tokens_details', 'cached_tokens'], dest='cache_read_tokens', required=False
                     ),
                     UsageExtractorMapping(
+                        path=['prompt_tokens_details', 'cache_write_tokens'], dest='cache_write_tokens', required=False
+                    ),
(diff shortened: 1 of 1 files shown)


Kludex:
Thanks for the PR. Closing it, since it can't merge as-is and the useful part has since landed elsewhere:

- `packages/python/genai_prices/data.py` is generated from `prices/providers/*.yml` (`make build`), so the mapping has to go in the provider YAML - the pre-commit build hook reverts hand edits, and `data.ts` on the JS side was not updated either. It also now conflicts with `main`.
- OpenAI's `prompt_tokens_details.cache_write_tokens` / `input_tokens_details.cache_write_tokens` mapping was added in #463, and the OpenAI-compatible copies (`huggingface_*`, `ovhcloud`, `doubleword`, `openrouter`, `modal`) carry it too.
- The remaining real gap is Azure OpenAI, which #596 covers with the YAML change, regenerated artifacts, a test and live request evidence.
- The other providers here (DeepSeek, Fireworks, Google, Moonshot, xAI) don't document that field; if one of them does emit it, a YAML change to that provider with a captured response would be welcome.

Issue #185, opened 2025-11-07
Support image tokens (input & output)


alexmojaki:
https://platform.openai.com/docs/pricing#image-tokens

<img width="732" height="297" alt="Image" src="https://github.com/user-attachments/assets/c090b8cb-d247-4eb6-beb4-348d50a441f6" />

## The gap

`ModelPrice` has text and audio token fields (`input_mtok`, `output_mtok`, `input_audio_mtok`, `output_audio_mtok`, `cache_*`) but **nothing for image tokens**, and `Usage` mirrors it (no `*_image_tokens`). So image-token charges — for both **input** image tokens (vision) and **output** image tokens (image generation) — currently have no representation, and multimodal costs are under-reported.

OpenAI now bills image generation per image token, e.g. (per 1M): `gpt-image-1-mini` output $8, `gpt-image-1.5` output $32, `gpt-image-2` output $30.

## Why it matters (concrete evidence)

In #427 we added `gpt-image-1-mini`, `gpt-image-1.5`, and `gpt-image-2`. Because there's no image field, the only options were to price text tokens and either drop image output or shoehorn it into `output_mtok`. That forced an inconsistency: `gpt-image-1.5` got a text-output guess (`output_mtok: 10`) while `gpt-image-1-mini` and `gpt-image-2` omit output entirely — none of which captures OpenAI's actual image-output pricing above. Any image-generation model added going forward hits the same wall.

## Extractability

Image-token counts are reported by the APIs, so cost calculation is feasible once the schema supports it:
- OpenAI returns a modality breakdown in `input_tokens_details` / `output_tokens_details`.
- Gemini returns per-modality `promptTokenDetails` (confirmed in #410, which includes an `IMAGE` modality token count).

## Relationship to the unit-registry work

This overlaps the arbitrary-unit direction in #410 and the data-driven unit registry stack (#356 → #351). If that lands, image tokens are just one instance of it — so this issue is best treated as the concrete OpenAI image-token case that the registry should cover, rather than a separate fixed-field design (`input_image_mtok` / `output_image_mtok`) that would cut against it.

Kludex:
Image-token pricing is now supported by the v2 unit registry published in [#512](https://github.com/pydantic/genai-prices/pull/512). Both engines support `input_image_tokens` and `output_image_tokens`, and current OpenAI and Google extractors and price entries use those units.

PR #351, opened 2026-05-01
Data-driven units phase 1

alexmojaki:
## Summary

- move token-unit metadata into the generated unit registry while preserving complete join closure
- add aggregate and per-modality output reasoning units
- extract documented reasoning usage from OpenAI, Azure, OpenRouter, Google, Groq, xAI, and Perplexity response shapes
- preserve distinct reasoning rates from OpenRouter and LiteLLM source catalogs
- model Perplexity Sonar Deep Research's published $3/MTok reasoning rate alongside its $8/MTok output rate
- price Google image-generation output with a text catch-all and explicit image child when modality details omit the text remainder

## Data-model decision

Publication still validates the complete registered dimension graph. That keeps every advertised price key usable by provider data or a custom `ModelPrice`, including required modality/reasoning joins. Per-model validation separately checks only the prices a selected model supplies. Unused registered units do not add per-calculation decomposition work.

A catch-all rate follows the provider's unclassified remainder, not the model's headline modality. For mixed Gemini image output, `output_mtok` is the text/thinking rate and `output_image_mtok` is the image rate. This prices omitted text correctly without fabricating an `output_text_tokens` report.

## Verification

- `uv run pytest -q` (771 passed, 41 xfailed)
- `uv run python tests/dataset/extract_usages.py`
- `npm test` (737 passed, 1 skipped)
- Python and TypeScript type checks via pre-commit


packages/js/src/engine.ts
alexmojaki, line 1: why has this file changed

alexmojaki: (Claude) It changed in `6a1bbba` to keep the shared JS pricing path aligned with the new Python registry decomposition for the generated parity fixtures. When overlapping priced units exist, the JS path now subtracts only separately priced descendants from a parent bucket, matching the Python behavior used to regenerate snapshots. No public JS API change was intended.

packages/python/genai_prices/types.py
```diff
@@ -188,82 +193,257 @@ def __radd__(self, other: ExtractedUsage | Any) -> ExtractedUsage:
...
 
-    @property
-    def cache_write_tokens(self) -> int | None:
-        """Number of tokens written to the cache."""
+    _values: dict[str, int]
 
-    @property
-    def cache_read_tokens(self) -> int | None:
-        """Number of tokens read from the cache.
+    def __init__(self, **kwargs: int | None) -> None:
```
alexmojaki, line 200: does this allow arbitrary kwargs? it should not, as long as it didn't before

packages/python/genai_prices/types.py
```diff
@@ -339,24 +519,13 @@ def summary(self) -> str:
...
-
-
 @dataclass
 class UsageExtractorMapping:
     """Mappings from used to build usage."""
 
     path: ExtractPath
     """Path to the value to extract"""
-    dest: UsageField
+    dest: str
```
alexmojaki, line 424: this should still be validated at runtime

packages/python/genai_prices/data.py
```diff
@@ -10407,3 +10408,38 @@
         ],
     ),
 ]
+unit_families_data: dict[str, Any] = {
```
alexmojaki: I believe that a feature of the library before this branch in main is that users can use their own provider data and avoid the cost of importing the built-in generated data if they want. If that's true, make sure that keeping that guarantee is enforced in the spec. That implies that this can't live here, because basic usage of the library currently requires getting the unit registry, which means, by default, importing it from here indirectly. Any attempt to use custom data would still end up importing this just to get the units.

I think that there needs to be a separate generated Python file containing the units data only. It will be a newly generated file and it will be separate from the provided data to make sure that importing it is cheap.

packages/python/genai_prices/types.py
```diff
@@ -509,6 +677,44 @@ def _type_name(v: Any) -> str:
...
+    from genai_prices.units import _get_registry  # pyright: ignore[reportPrivateUsage]
+
+    return _get_registry().reported_usage_keys()
+
+
+def _reported_usage_key_order() -> tuple[str, ...]:
+    from genai_prices.units import _get_registry  # pyright: ignore[reportPrivateUsage]
+
+    registry_keys = tuple(_get_registry().reported_usage_keys())
+    extra_keys = tuple(key for key in registry_keys if key not in _USAGE_REPR_ORDER)
```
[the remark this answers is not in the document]
alexmojaki: (Claude) Addressed in 8d33283

packages/python/genai_prices/types.py
alexmojaki: (Claude) Already covered: runtime extraction returns Usage(**values), so destination strings are validated against the active registry when the extractor runs. tests/test_extract_usage.py includes test_extractor_rejects_invalid_destination_string for this path.

packages/python/genai_prices/types.py
alexmojaki: (Claude) Addressed in 6ee43ce

packages/python/genai_prices/types.py
```diff
@@ -509,6 +677,44 @@ def _type_name(v: Any) -> str:
     return 'None' if v is None else type(v).__name__
 
 
+_USAGE_REPR_ORDER = (
```
alexmojaki: if this exists only to preserve legacy repr and keep old tests passing as is, remove it and only use other deterministic ordering

packages/python/genai_prices/types.py
```diff
@@ -509,6 +677,44 @@ def _type_name(v: Any) -> str:
...
+    'cache_write_tokens',
+    'cache_read_tokens',
+    'output_tokens',
+    'input_audio_tokens',
+    'cache_audio_read_tokens',
+    'output_audio_tokens',
+)
+
+
+def _reported_usage_keys() -> frozenset[str]:
```
alexmojaki, line 605: add a note here and in phase 5 spec to look at improving performance around here

packages/python/genai_prices/data.py
alexmojaki: (Claude) Addressed in `a02f244` and `d57d86b`: the phase-one spec now requires separate cheap Python unit data, and the generator now writes `data_units.py` for `unit_families_data` while `data.py` exports only providers. `tests/test_lifecycle.py` includes a subprocess check that importing `genai_prices.data_units` does not import `genai_prices.data`.

packages/python/genai_prices/types.py
alexmojaki: (Claude) Addressed in `6d3e176` and `09da0c4`: the phase-one spec now says `Usage.__repr__` uses registry unit order, and the hardcoded legacy `_USAGE_REPR_ORDER` tuple has been removed. The test now asserts default registry order instead of legacy dataclass field order.

specs/data-driven-unit-registry/phase-1-python-internal-registry/spec.md
```diff
@@ -17,8 +17,11 @@ The Phase 1 registry exposes only the usage keys and price keys Python already s
 **The remote payload shape matters because prices update outside package releases.** _(from "Phase 1 preserves supported entry points while changing unsafe internals")_
 The packages load generated language-native data at startup, but long-running clients can fetch `prices/data.json` or `prices/data_slim.json` to receive updated prices before a package release. If Phase 1 changed those remote files to include unit families, older released clients would break. Phase 1 therefore keeps both remote payloads as provider arrays and embeds only the current-unit registry subset into Python generated package data.
 
+**Python unit data stays separate from generated provider data.** _(from "The remote payload shape matters because prices update outside package releases", "Phase 1 preserves supported entry points while changing unsafe internals")_
```
alexmojaki: make the JS spec match

alexmojaki: (AI) Addressed in 8d4ebd0

packages/python/genai_prices/types.py
alexmojaki: it should be validated when the extractor is constructed

prices/src/prices/prices_types.py
```diff
@@ -201,7 +190,7 @@ def is_match(self, model_id: str) -> bool:
 
     @field_validator('prices_checked', mode='after')
     @classmethod
-    def validate_prices_checked(cls, prices_checked: date | None, info: FieldValidationInfo) -> date | None:
```
alexmojaki: why did this change

prices/src/prices/prices_types.py
```diff
@@ -121,23 +121,12 @@ def exclude_free(self):
...
-]
-
-
 class UsageExtractorMapping(_Model):
     """Mappings from used to build usage."""
 
     path: ExtractPath
     """Path to the value to extract"""
-    dest: UsageField
+    dest: str
```
alexmojaki, line 132: let's leave this unchanged in the build pipeline for now, it doesn't seem to be proving anything and it's gonna change again in phase 4 anyway, right?

packages/python/genai_prices/types.py
alexmojaki: (AI) Addressed in 2c54586

prices/src/prices/prices_types.py
alexmojaki: (AI) Kept intentionally: `FieldValidationInfo` now emits a Pydantic deprecation warning under the current dependency set, and this repo treats warnings as errors during pytest collection. `ValidationInfo` is the non-deprecated replacement. I reverted the unrelated build extractor destination/schema change in 74790ba.

prices/src/prices/prices_types.py
alexmojaki: (AI) Addressed in 74790ba

packages/python/genai_prices/types.py
```diff
@@ -610,104 +837,36 @@ class ModelPrice:
...
-            raise ValueError('output_audio_tokens cannot be greater than output_tokens')
-
-        output_price += calc_mtok_price(self.output_mtok, priced_text_output_tokens, total_input_tokens)
-        output_price += calc_mtok_price(self.output_audio_mtok, usage.output_audio_tokens, total_input_tokens)
-
-        total_price = input_price + output_price
+        input_price = Decimal(0)
+        output_price = Decimal(0)
+        total_price = Decimal(0)
+        total_input_tokens = usage_data.input_tokens if _model_price_uses_tiered_prices(self, registry) else 0
```
alexmojaki: why the conditional

prices/src/prices/prices_types.py
alexmojaki: make sure the spec matches

prices/src/prices/prices_types.py
alexmojaki: why would that only surface in this pr

packages/python/genai_prices/types.py
alexmojaki: (AI) Addressed in faf36b9

prices/src/prices/prices_types.py
alexmojaki: (AI) Addressed in 0aca511 and 74790ba: the Phase 1 code spec now says the build-time `UsageField`, `UsageExtractorMapping.dest` annotation, and generated schema enum stay unchanged until the later registry-derived schema/autocomplete phase, and the build pipeline/schema were restored accordingly.

prices/src/prices/prices_types.py
alexmojaki: (AI) It surfaced here because this PR added Phase 1 tests that import `prices.package_data` / `prices.prices_types` under pytest warnings-as-errors. With the current Pydantic version, importing `FieldValidationInfo` emits a deprecation warning during collection, so those new tests fail before running. `ValidationInfo` is the non-deprecated equivalent; the build extractor destination/schema change was the unrelated part and was reverted in 74790ba.

packages/python/genai_prices/types.py
alexmojaki: under what conditions would an error not be raised anyway? i'm trying to see if there's a quirk in the spec

alexmojaki: (AI) The no-error case is when `input_tokens` is underdetermined but irrelevant to the selected price. Example: `Usage(input_audio_tokens=200_000, cache_read_tokens=50_000, output_tokens=10_000)` cannot infer a unique `input_tokens`, but `ModelPrice(output_mtok=Decimal("1"))` only prices `output_tokens` and has no tier threshold, so pricing can ignore the inconsistent input-side reports. If the selected model price uses a tiered value, or prices an input-side unit that requires decomposing/infering `input_tokens`, then the error is still raised. This matches the spec line that unpriced reported keys are ignored unless needed to infer a missing priced value or missing tier threshold; `tests/test_tiered_pricing.py::test_non_tiered_price_does_not_read_unneeded_input_token_threshold` covers it.

prices/src/prices/prices_types.py
alexmojaki: is it fair to say that the change here wasn't achieving anything

alexmojaki: (AI) Yes. For Phase 1, changing the build-time `UsageExtractorMapping.dest` from the hardcoded `UsageField` enum to `str` was not buying useful behavior: current provider YAML still only uses current known destinations, package-data validation already checks those after parsing, and registry-derived provider schema/autocomplete is Phase 4 work. The meaningful Phase 1 change is only on the runtime package side, where `packages/python/genai_prices/types.py` accepts string destinations and validates runtime-created extractors against the active registry.

prices/src/prices/prices_types.py
alexmojaki: nothing tested the build pipeline before?

alexmojaki: (AI) The build pipeline was exercised by the build command/hooks, but not in a pytest collection path where warnings are configured as errors. This PR added tests that import `prices.package_data` / `prices.prices_types` directly under pytest, so the `FieldValidationInfo` deprecation warning became a collection error. Running the build command itself only emits/ignores that warning path rather than failing the way pytest does.

packages/python/genai_prices/types.py
alexmojaki: where in the spec

alexmojaki: (AI) In `specs/data-driven-unit-registry/spec.md`: lines 80-81 say only priced units become buckets and "Unpriced reported usage keys are ignored unless they are needed to infer a missing priced value or a missing tier threshold"; lines 89-90 say tier selection uses the provided-or-inferable `input_tokens` total and raises only when that threshold is needed but cannot be inferred coherently; lines 92-93 state the lazy rule that contradictory/underdetermined data raises only when that missing value or a priced bucket is needed.

The Phase 1 implementation spec is even more direct: `specs/data-driven-unit-registry/phase-1-python-internal-registry/code-spec.md` lines 218-225 list the pricing flow step "read `input_tokens` only when tiered pricing needs a threshold", and lines 232-243 say non-tiered prices get a neutral threshold because they ignore it.

alexmojaki: (AI) Correction to the prior reply with the identifiers intact: `specs/data-driven-unit-registry/spec.md` lines 80-81 say unpriced reported usage keys are ignored unless needed to infer a missing priced value or a missing tier threshold. Lines 89-90 define the tier threshold as the provided-or-inferable `input_tokens` total and say pricing raises only when that threshold is needed but cannot be inferred coherently. Lines 92-93 state the same lazy rule for contradictory or underdetermined usage.

The Phase 1 code spec is explicit at `specs/data-driven-unit-registry/phase-1-python-internal-registry/code-spec.md`: line 224 says to read `input_tokens` only when tiered pricing needs a threshold, and line 243 says to pass a neutral threshold when no configured price uses `TieredPrices` because non-tiered prices ignore it.

packages/python/genai_prices/types.py
```diff
@@ -398,17 +572,41 @@ def extract(self, response_data: Any) -> tuple[str | None, Usage]:
...
+    from genai_prices.validation import validate_extractor_destinations
+
+    validate_extractor_destinations(
+        {mapping.dest for mapping in mappings},
+        _reported_usage_keys_for_extractor_construction(),
+    )
+
+
+def _reported_usage_keys_for_extractor_construction() -> frozenset[str]:
+    data_module = sys.modules.get('genai_prices.data')
```
alexmojaki: why do we need this hack? if there's a reason, add a comment and a note to revisit in phase 4 or 5

alexmojaki: (AI) Addressed in 3ea29d7

packages/python/genai_prices/units.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @staticmethod
+    def are_compatible(a: UnitDef, b: UnitDef) -> bool:
+        """Return whether two units can overlap without conflicting dimensions."""
+        if a.family is not b.family:
+            return False
+
+        return all(b.dimensions.get(key, value) == value for key, value in a.dimensions.items())
+
+    @staticmethod
```
alexmojaki: no public staticmethods, it increases API surface. move to functions where sensible. even if they remain public, at least they'll be more tucked away in some module. in this particular case decompose.is_descendant_or_self seems to provide the same value. and even if it didn't, what about a normal method on UnitDef? could do that with e.g. are_compatible, probably others. update specs accordingly.

alexmojaki: (AI) Addressed in 8c754e1. The spec update is kept in this PR, not in the spec-base branch.

packages/python/genai_prices/units.py
```diff
@@ -0,0 +1,147 @@
...
+    per: int
+    description: str
+    units: dict[str, UnitDef] = field(default_factory=dict)
+
+
+class UnitRegistry:
+    families: dict[str, UnitFamily]
+    units: dict[str, UnitDef]
+    price_keys: dict[str, str]
+    _units_by_dimension: dict[str, dict[frozenset[tuple[str, str]], UnitDef]]
```
alexmojaki: wouldn't this make more sense on the family?

alexmojaki: (AI) Addressed in a523cb6. The dimension index now lives on UnitFamily as units_by_dimension, with the corresponding spec update kept in this PR.

packages/python/genai_prices/units.py
```diff
@@ -0,0 +1,147 @@
...
+    id: str
+    per: int
+    description: str
+    units: dict[str, UnitDef] = field(default_factory=dict)
+
+
+class UnitRegistry:
+    families: dict[str, UnitFamily]
+    units: dict[str, UnitDef]
+    price_keys: dict[str, str]
```
alexmojaki: it looks like most of the time we want the unit, not just the unit id (usage key). also this is for performance, right? exposing it publicly seems bad. would a private field and a method to get unit by price key make more sense

alexmojaki: (AI) Addressed in c3fd88a. The registry now keeps a private _units_by_price_key index and exposes unit_for_price_key() for callers that need the UnitDef.

packages/python/genai_prices/units.py
```diff
@@ -0,0 +1,144 @@
...
+        for usage_key, unit in self.units.items():
+            self._ancestor_usage_keys[usage_key] = frozenset(
+                maybe_ancestor.usage_key
+                for maybe_ancestor in unit.family.units.values()
+                if maybe_ancestor is not unit and _is_dimension_subset(maybe_ancestor, unit)
+            )
+
+        self._validate_interval_closure()
+
+    def find_join(self, a: UnitDef, b: UnitDef) -> UnitDef | None:
```
alexmojaki: why is this method on the registry

packages/python/genai_prices/units.py
```diff
@@ -0,0 +1,144 @@
...
+                                    f'{key}={value}' for key, value in sorted(required_dimensions)
+                                )
+                                raise ValueError(
+                                    f'Missing intermediate unit dimensions in family {family.id} '
+                                    f'between {ancestor.usage_key} and {descendant.usage_key}: '
+                                    f'{missing_dimensions}'
+                                )
+
+
+def are_compatible(a: UnitDef, b: UnitDef) -> bool:
```
alexmojaki: why not a method on unitdef

packages/python/genai_prices/validation.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import annotations
+
+from genai_prices.units import UnitDef, UnitFamily, UnitRegistry, are_compatible
+
+__all__ = (
```
alexmojaki: why have an all here, would users have any reason to import these

alexmojaki: similar question for other new niche modules

packages/python/genai_prices/units.py
alexmojaki: (AI) Addressed in 38246d9. Join lookup now lives on UnitFamily, which owns the dimension index it uses.

packages/python/genai_prices/validation.py
alexmojaki: (AI) Addressed in ea506ed. Removed __all__ from the new helper modules validation.py, decompose.py, and units.py; they are implementation modules, not curated public surfaces.

packages/python/genai_prices/units.py
alexmojaki: (AI) Addressed in ea506ed. Compatibility is now UnitDef.is_compatible_with().

packages/python/genai_prices/types.py
```diff
@@ -509,6 +605,35 @@ def _type_name(v: Any) -> str:
...
+    from genai_prices.units import _get_registry  # pyright: ignore[reportPrivateUsage]
+
+    registry = _get_registry()
+    reported_keys = registry.reported_usage_keys()
+    return tuple(key for key in registry.units if key in reported_keys)
+
+
+def _raw_usage_value(obj: object, key: str) -> int | None:
+    if isinstance(obj, Mapping):
+        value = cast(Mapping[str, object], obj).get(key)
```
alexmojaki: i thought this was dropped from the spec. what does it say?

alexmojaki: (AI) The spec still includes this. Phase 1 says `Usage.from_raw(obj)` reads known externally reported usage keys from mappings or objects and ignores extras; the broader spec also says callers can pass mappings, dataclasses, namespace objects, or other objects as usage. This line is the mapping half of that raw-object path, not direct `Usage(...)` construction.

alexmojaki: (AI) Corrected in ade72c7 and d12abbc. The spec now says mapping usage inputs are not part of the Python calc_price compatibility surface, and Usage.from_raw no longer reads values from mappings.

dsfaccini:
David's AICA here: we want to unblock image-token pricing (#185), and since the right home for it is this unit-registry work, I did a trial rebase of this PR onto current `main` to gauge what reviving it would take. Sharing the results in case they help you pick it back up (or hand it off).

**Good news — the expensive risk is retired.** Rebased onto current `main` (the ~47 commits since mid-May), it merges, builds, lints, and typechecks (Python + TS), with **682/689 tests passing**. Crucially, **all of the new provider data added since May passes the registry's structural-closure / ancestor-coverage validation with zero errors** — the "registry rejects two months of new data" scenario didn't happen.

Pushed the rebased branch as [`trial-revive-351`](https://github.com/pydantic/genai-prices/compare/main...trial-revive-351) (2 commits on current `main`) so you can inspect it directly — it's a trial artifact, not a PR.

**Two items that need your call** (the real remaining cost):

1. **Price serialization** — post-merge, the build bypasses `serialize_decimal`, so whole-number prices render as `2.0` instead of `2` in `data.json`/`data.py` (bloats every generated diff, breaks 3 CLI snapshots). Looks like an interaction between the registry's price/conditional-price handling and main's newer `build.py` / `export_validation.py` export path — which path is intended?
2. **Overlap vs ancestor-coverage validation** — the registry swapped main's overlap checks for ancestor-coverage, so 4 of main's `test_model_price_rejects_impossible_overlapping_usage` cases now hit the registry's different `ValueError` (impossible usage is still rejected, just a different message/contract). Port the overlap checks into the registry model, or rewrite those tests to the new contract?

Happy to take the mechanical landing work off your plate (rebase + test reconciliation) if the design's still the direction — just say the word. The motivation is #185: once this lands, image-token pricing becomes a data-only addition rather than a fixed-field schema change.

dsfaccini:
David's AICA here: following up — rather than leave it on a side branch, David had me push the rebase **directly onto this PR branch** (fast-forward from `a90aed3`, so nothing was rewritten — you can drop the two commits with a reset if you don't want them). The standalone `trial-revive-351` branch still exists too.

**The two commits to actually review are ours:**
- [`cd695c9`](https://github.com/pydantic/genai-prices/commit/cd695c9) — the `main` merge: all 8 conflict resolutions + the semantic test reconciliation (registry-lifecycle tests unioned with main's new concurrency tests) + the `httpx`→`httpx2` fixup.
- [`02c263a`](https://github.com/pydantic/genai-prices/commit/02c263a) — regenerated `test_extract_usage` snapshots for the registry's modality sub-tokens.

**Heads up on the diff size:** this PR targets `codex/data-driven-unit-registry-spec-base`, which is also behind `main`, so pulling `main` in here makes the PR now show all ~47 of main's recent commits. To get a clean diff you'd want to either merge `main` into the `-spec-base` branch too, or retarget this PR to `main`. Nothing's wrong — it's just the stacked-base showing through.

The two open design questions from my previous comment (price `serialize_decimal` bypass → `2.0`; overlap-check vs ancestor-coverage validation) are unchanged and still need your call. Feel free to point your own Claude at `cd695c9` to sanity-check the resolutions align with the registry's intent — happy to iterate.

alexmojaki:
@cubic-ai-dev review

packages/js/src/units.ts
```diff
@@ -0,0 +1,124 @@
...
+export function setActiveRegistry(registry: null | UnitRegistry): void {
+  if (registry === null) {
+    activeRegistry = generatedRegistry
+  } else {
+    activeRegistry = registry
+  }
+}
+
+export function dimensionKey(dimensions: Record<string, string>): string {
+  return Object.entries(dimensions)
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.
(David assigned this triage task but has not separately reviewed this reply.)

Fixed in 1dd32a2. `dimensionKey` now encodes the sorted entries structurally via `JSON.stringify`, matching the Python registry's `frozenset(dimensions.items())` key. This removes the unescaped-delimiter collision path — distinct dimension sets can no longer map to the same string.

packages/js/src/__tests__/decompose.test.ts
```diff
@@ -0,0 +1,164 @@
...
+        }),
+        getActiveRegistry().units
+      )
+    ).toThrow(
+      'Invalid usage data: more-specific usage for cache_read_tokens, input_audio_tokens totals 160, which exceeds input_tokens (100)'
+    )
+  })
+})
+
+describe('Python Phase 1 decomposition parity examples', () => {
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.
(David assigned this triage task but has not separately reviewed this reply.)

Fixed in 1dd32a2. Removed the `Python Phase 1 decomposition parity examples` block — its two assertions duplicated the `parent/child` and `cached-audio overlap` cases already covered in the `computeLeafValues` suite, so no coverage was lost.

prices/src/prices/export_validation.py
```diff
@@ -0,0 +1,160 @@
...
+        'this',
+        'throw',
+        'true',
+        'try',
+        'typeof',
+        'var',
+        'void',
+        'while',
+        'with',
+        'yield',
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.
(David assigned this triage task but has not separately reviewed this reply.)

Fixed in 1dd32a2. Added the strict-mode/future reserved words `let`, `static`, `implements`, `interface`, `package`, `private`, `protected`, `public`, plus the strict-mode-restricted binding names `arguments` and `eval`, to `_JAVASCRIPT_KEYWORDS`.

packages/js/src/decompose.ts
```diff
@@ -0,0 +1,57 @@
...
+  for (const unit of pricedUnits) {
+    let leafValue = 0
+    for (const descendant of pricedUnits) {
+      if (!isDescendantOrSelf(unit, descendant)) continue
+
+      const sign = (Object.keys(descendant.dimensions).length - Object.keys(unit.dimensions).length) % 2 === 0 ? 1 : -1
+      leafValue += sign * getUsageValue(usage, descendant.usageKey)
+    }
+
+    if (leafValue < 0) {
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.
(David assigned this triage task but has not separately reviewed this reply.)

Not applying this — it would diverge from the Python reference implementation, which is the source of truth for this port. Usage values are integer token counts by contract: Python stores them as `dict[str, int]` and `compute_leaf_values` returns `dict[str, int]` with exact integer comparison and no tolerance. Fractional token usage doesn't occur, so a floating-point residue can't arise, and adding a tolerance clamp to the JS side only would break parity with Python. `normalizeUsage`/`getUsageValue` already reject non-finite and negative values. Resolving as by-design.

packages/js/src/usage.ts
```diff
@@ -0,0 +1,101 @@
...
+
+function validateUsageValue(usageKey: string, value: number): number {
+  if (!Number.isFinite(value) || value < 0) {
+    throw new Error(`Invalid usage value for ${usageKey}: expected a finite non-negative number`)
+  }
+  return value
+}
+
+function isPlainObject(obj: unknown): obj is Record<string, unknown> {
+  return typeof obj === 'object' && obj !== null && !Array.isArray(obj)
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.
(David assigned this triage task but has not separately reviewed this reply.)

Not applying this — the permissiveness is intentional and matches the Python reference: `Usage.from_raw` reads each reported key via `getattr(obj, key, None)` on *any* object, so class instances and other object types are supported by design (this is how usage objects from provider SDKs get read). The `!Array.isArray` check already excludes arrays, and `normalizeUsage` only keeps entries where `typeof value === 'number'`. Restricting to plain objects only would diverge from Python and drop legitimate usage objects. Resolving as by-design.

packages/python/genai_prices/types.py
```diff
@@ -747,22 +785,99 @@ def calc_mtok_price(
...
+        bad_keys = ', '.join(sorted(unknown_price_keys))
+        raise ValueError(f'Unknown price key: {bad_keys}')
+
+    return tuple(
+        (registry.unit_for_price_key(price_key), cast(Decimal | TieredPrices, value))
+        for price_key, value in stored_prices
+    )
+
+
+def _collect_model_price_units(  # pyright: ignore[reportUnusedFunction]
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in ec8fabb

packages/python/genai_prices/types.py
```diff
@@ -747,22 +785,99 @@ def calc_mtok_price(
...
-        price = applicable_price * token_count
+        unit_price = applicable_price * count
     else:
-        price = field_mtok * token_count
-    return price / 1_000_000
+        unit_price = price * count
+    return unit_price / per
+
+
+def _collect_effective_model_price_keys(  # pyright: ignore[reportUnusedFunction]
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 961e57f

packages/python/genai_prices/units.py
```diff
@@ -0,0 +1,99 @@
...
+    per: int
+    dimensions: dict[str, str]
+
+    def is_compatible_with(self, other: UnitDef) -> bool:
+        """Return whether two units can overlap without conflicting dimensions."""
+        return all(other.dimensions.get(key, value) == value for key, value in self.dimensions.items())
+
+
+class UnitRegistry:
+    units: dict[str, UnitDef]
```
alexmojaki: make this immutable

packages/python/genai_prices/units.py
```diff
@@ -0,0 +1,99 @@
...
+    all_usage_keys: frozenset[str]
+    all_price_keys: frozenset[str]
+    reported_usage_keys: frozenset[str]
+    reported_usage_keys_in_order: tuple[str, ...]
```
alexmojaki: make these protected

packages/js/src/units.ts
```diff
@@ -0,0 +1,112 @@
+import type { RawUnitsDict, UnitDef } from './types'
+
+import { unitData } from './dataUnits'
+
+export class UnitRegistry {
```
alexmojaki, line 5: make this immutable and minimise api surface as with python, update spec to match. prevent misuse.

packages/python/genai_prices/units.py
alexmojaki: (AI) Addressed in 9990705

packages/python/genai_prices/units.py
alexmojaki: (AI) Addressed in 3b3f1a5

packages/js/src/units.ts
alexmojaki: (AI) Addressed in 6648c26

dsfaccini:
This comment was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Alex, I’ve caught up through `f4c334c`. The registry-immutability and reasoning-lattice work resolves the earlier concerns and is a substantial improvement. I found one remaining provider-contract problem that currently makes an official response unpriceable and incomplete.

1. **Normalize Perplexity reasoning before treating it as an output descendant.** Perplexity’s [official Deep Research response](https://docs.perplexity.ai/docs/sonar/models/sonar-deep-research) reports `completion_tokens=11,395`, `reasoning_tokens=193,947`, and `total_tokens=11,428`; reasoning is additional rather than contained in completion. With the [current extractor](https://github.com/pydantic/genai-prices/blob/f4c334c64c2398c2d6c2d889bc375fe55192b40c/prices/providers/perplexity.yml), public `extract_usage(...).calc_price()` raises:

   `Invalid usage data: output_reasoning_tokens (193947) cannot exceed output_tokens (11395)`

   Because extractor destinations accumulate, the smallest registry-compatible normalization is to map `reasoning_tokens` to both `output_tokens` and `output_reasoning_tokens`, retaining `completion_tokens -> output_tokens`. That produces inclusive `output_tokens=205,342`, while decomposition charges the 11,395 completion tokens at the ordinary output rate and 193,947 reasoning tokens at the reasoning rate.

2. **Complete Deep Research’s billable surface before the v2 vocabulary freezes.** [Perplexity separately bills](https://docs.perplexity.ai/docs/getting-started/pricing) `citation_tokens` at $2/MTok and `num_search_queries` at $5/K, but the current extractor and model prices omit both.

   - Model citation tokens as a separate reportable family with `direction: output`; the official citation count exceeds completion, so it should not be an `output_tokens` descendant.
   - Model search queries as a separate reportable family. Don’t reuse `requests`, which the [registry contract](https://github.com/pydantic/genai-prices/blob/f4c334c64c2398c2d6c2d889bc375fe55192b40c/specs/data-driven-unit-registry/spec.md) reserves for the pricing-supplied count of one API request and forbids as an extractor destination.
   - Add their extractor mappings and Deep Research price keys/rates.

3. **Replace the provider-specific synthetic assertions with the official end-to-end contract.** Through public Python `extract_usage(...).calc_price()` and equivalent JS `extractUsage` + `calcPrice`, assert:

   - normalized usage, including reasoning, citations, and search queries;
   - `input_price == 0.000066`;
   - `output_price == 0.711057`;
   - `total_price == 0.816123`.

   This replaces the current Perplexity extraction fixture, which encodes reasoning as a subset of completion, and the hand-built provider pricing test that bypasses extraction. I’d retain the custom-`ModelPrice` test as the provider-independent decomposition test.

prices/providers/perplexity.yml
```diff
@@ -2,11 +2,25 @@
...
 price_comments: >-
   Prices per request vary based on usage, this is not represented here,
   instead we just take the highest price shown for `requests_kcount`.
+extractors:
+  # https://docs.perplexity.ai/api-reference/sonar-post
+  - root: usage
+    mappings:
+      - path: prompt_tokens
+        dest: input_tokens
+      - path: reasoning_tokens
```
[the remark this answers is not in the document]
alexmojaki: (AI) Thanks, no action needed. This is provider-specific normalization backed by Perplexity's documented Deep Research response: `prompt_tokens=33`, `completion_tokens=11395`, and `total_tokens=11428`, while `reasoning_tokens=193947` is reported separately. Because `total_tokens` is exactly prompt plus completion, Perplexity's reasoning count is not a subset of completion in this response. The end-to-end regression test preserves those exact values and expects normalized output to include completion, reasoning, and citation tokens.

packages/js/src/data.ts
```diff
@@ -10603,11 +10948,115 @@ export const data: Provider[] = [
...
+          },
+          {
+            path: 'output_tokens',
+            dest: 'output_tokens',
+            required: true,
+          },
+        ],
+      },
+      {
+        api_flavor: 'realtime',
```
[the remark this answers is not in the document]
alexmojaki: (AI) Thanks, no action needed. Model-less extraction is an existing supported state: the extractor returns the usage with `model=None` when a response schema omits a model. Python callers can supply `model` to `ExtractedUsage.calc_price`, while JS `calcPrice` already takes the model separately. The Realtime extractor is intentionally useful for extracting the event's modality and cache usage without claiming that `response.done` is self-contained.

tests/test_end_to_end.py
```diff
@@ -75,6 +75,33 @@ def test_anthropic_caching_read():
...
+    extracted_usage = extract_usage(response, provider_id='anthropic')
+    assert extracted_usage.usage == Usage(
+        input_tokens=100,
+        cache_write_tokens=0,
+        cache_read_tokens=0,
+        output_tokens=50,
+        web_searches=2,
+    )
+    price = extracted_usage.calc_price()
+    assert price.input_price == Decimal('0.0003')
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 68e8ca1

prices/providers/perplexity.yml
```diff
@@ -67,9 +90,16 @@ models:
...
+    prices_checked: 2026-07-21
+    price_comments: >-
+      Perplexity lists reasoning tokens separately at $3 per million tokens and citation tokens at $2 per million tokens.
+      Search queries cost $5 per thousand.
     prices:
       input_mtok: 2
       output_mtok: 8
+      output_reasoning_mtok: 3
+      output_citation_mtok: 2
+      web_searches_kcount: 5
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 8e2106e

packages/js/src/__tests__/dataset.test.ts
```diff
@@ -97,13 +98,16 @@ describe('dataset', () => {
             if (price) {
               expect(price.input_price).toBeCloseTo(parseFloat(extractor.input_price!), 8)
               expect(price.output_price).toBeCloseTo(parseFloat(extractor.output_price!), 8)
+              if (extractor.total_price !== undefined) {
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 938f719

tests/test_cli.py
```diff
@@ -422,7 +493,187 @@ def test_calc_table_split_columns_rich_prices(monkeypatch: pytest.MonkeyPatch, c
     assert cli_logic(['calc', '--table', '--input-tokens', '1000', tiered_ref, requests_ref]) == 0
     out, err = capsys.readouterr()
     assert '(+tiers)' in out
-    assert 'Requests/K' in out
+    assert 'requests' in out.lower()
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in f338a85

PR #352, opened 2026-05-01
Add main-passing regression tests

alexmojaki:
Extracts the tests from #351 that are independent of the unit-registry implementation and already pass when run directly on `main`.

This PR is test-only. It keeps the behavior-changing assertions in #351, including the unpriced-descendant pricing regression and registry-specific export/update checks.

Included here:
- existing price-calculation parity regressions
- extractor accumulation, unknown-extra, contradiction-storage, and public extraction parity regressions
- provider-array remote payload and update-fetch parsing regressions

Verification:
- `uv run pytest tests/test_price_regressions.py tests/test_extract_usage.py tests/test_lifecycle.py tests/test_update_prices.py`
- `make test`
- `uv run basedpyright`


tests/test_price_regressions.py
```diff
@@ -0,0 +1,147 @@
+from decimal import Decimal
+
+import pytest
+
+from genai_prices import Usage
+from genai_prices.types import ModelPrice
+
+pytestmark = pytest.mark.anyio
```
alexmojaki: why

tests/test_price_regressions.py
```diff
@@ -0,0 +1,147 @@
...
+def test_model_price_decomposition_matches_current_text_cache_pricing() -> None:
+    price = ModelPrice(
+        input_mtok=Decimal('2'),
+        cache_write_mtok=Decimal('0.5'),
+        cache_read_mtok=Decimal('0.25'),
+        output_mtok=Decimal('10'),
+    ).calc_price(Usage(input_tokens=1_000, cache_write_tokens=20, cache_read_tokens=30, output_tokens=100))
+
+    assert price == {
+        'input_price': Decimal('0.0019175'),
```
alexmojaki: all these assertions need to make it clear where these numbers come from for this test to be readable and useful

tests/test_public_exports.py
alexmojaki, line 1: remove

tests/test_price_regressions.py
alexmojaki: (Claude) Addressed in c45e7b6

tests/test_price_regressions.py
alexmojaki: (Claude) Addressed in c45e7b6

tests/test_public_exports.py
alexmojaki: (Claude) Addressed in c45e7b6

PR #353, opened 2026-05-01
Fix audio pricing data

alexmojaki:
Stacks on #352. This is a data-only extraction from #351; no pricing engine or build logic changes are included.

Changes:
- Add Gemini cached-audio read pricing where cache-read and audio rates already imply it.
- Add matching text input/output prices for OpenAI audio/TTS model entries that previously only had audio-specific prices.
- Regenerate Python, TypeScript, and JSON provider data.
- Update the extracted usage dataset prices affected by those data fixes.

Verification:
- `make build-prices package-data`
- `make test`
- `uv run basedpyright`
- `npm run ci`


prices/providers/google.yml
```diff
@@ -326,6 +326,7 @@ models:
         tiers:
```
alexmojaki: remove this incorrect tier

prices/providers/openai.yml
```diff
@@ -371,6 +371,7 @@ models:
     context_window: 128000
     prices_checked: 2025-07-22
     prices:
+      input_mtok: 2.5
```
alexmojaki, line 376: add comments explaining these duplications

prices/providers/google.yml
alexmojaki: (Claude) Addressed in 8f37b37

prices/providers/openai.yml
alexmojaki: (Claude) Addressed in 8f37b37

alexmojaki: that's the wrong explanation

alexmojaki: (Claude) Addressed in 766b32b

PR #357, opened 2026-05-05
Defer usage inference to phase 8

alexmojaki:
## Summary

- moves missing-usage inference out of Phases 1-7 and into a new Phase 8 spec/code spec
- updates Phase 1/2 specs to use explicit-only missing usage behavior for reads and pricing
- clarifies direct missing reads raise when answering would require inferring an omitted ancestor or overlap

## Notes

This is a spec-only follow-up against the Phase 1 branch.


specs/data-driven-unit-registry/phase-1-python-internal-registry/code-spec.md
```diff
@@ -170,8 +178,8 @@ class Usage:
     def from_raw(cls, obj: object) -> Usage:
         """Wrap arbitrary usage input while ignoring unknown raw-object extras."""
 
-    def __getattr__(self, name: str) -> int:
-        """Return a stored registered value, lazily infer it, or raise a user-facing error."""
+    def __getattr__(self, name: str) -> int | None:
+        """Return a stored registered value, None for an unambiguous missing value, or raise."""
```
alexmojaki: why has this changed to returning None, don't we spec 0 elsewhere

specs/data-driven-unit-registry/phase-1-python-internal-registry/code-spec.md
```diff
@@ -182,9 +190,19 @@ class Usage:
...
+Direct construction is strict for registered externally reported usage keys and rejects unknown keyword names and non-reported pricing-only keys such as `requests`. `from_raw(...)` reads known externally reported usage keys from mappings or objects and ignores extras. Construction stores reported values only; it does not infer ancestors, normalize values, remember explicit-versus-inferred provenance, or reject contradictory registered values.
+
+For registered attribute reads:
+
+1. If the value was reported, return it directly without auditing descendants.
```
alexmojaki, line 191: does this need to duplicate the pricing part. wouldn't the pricing implicitly raise simply by going through this path.

alexmojaki: (AI) Addressed in 3543eba. I clarified that decomposition owns pricing missing-usage checks because pricing and direct `Usage` reads intentionally have different safe-missing semantics.

specs/data-driven-unit-registry/phase-1-python-internal-registry/code-spec.md
alexmojaki: (AI) Addressed in 3543eba. I clarified that Python direct reads return `None` for unambiguous missing values; the zero rule is pricing-only for safely absent priced units.

specs/data-driven-unit-registry/phase-1-python-internal-registry/code-spec.md
alexmojaki: and now?

specs/data-driven-unit-registry/phase-1-python-internal-registry/code-spec.md
alexmojaki: (AI) Updated in 55f7867. We are keeping the earlier zero behavior: safe missing direct reads return `0` without materializing that value as reported usage.

specs/data-driven-unit-registry/phase-1-python-internal-registry/code-spec.md
alexmojaki: (AI) Updated in 55f7867. The spec now says both direct reads and pricing return zero for safe missing values, but decomposition still owns pricing-specific missing ancestor/join checks because it can ignore unpriced descendants when an explicit priced ancestor is available.

alexmojaki: (AI) Correction in 5f8c225. You were right: decomposition does not need to duplicate missing ancestor/join checks. It reads only the selected priced keys through `Usage`, so `Usage` owns those missing-read errors; decomposition still owns only selecting priced keys and rejecting impossible negative exclusive buckets after reads.

PR #358, opened 2026-05-06
Implement Phase 2 JavaScript unit registry

alexmojaki:
## Summary

Implements Phase 2 of the data-driven unit registry for the JavaScript package.

This keeps Phase 2 scoped to JavaScript internals and packaging support:
- generated JavaScript unit-family data now lives in `packages/js/src/dataUnits.ts`, separate from provider-heavy `data.ts`
- JavaScript usage, model price, and extractor types are widened for registry-defined keys
- `units.ts`, `usage.ts`, `decompose.ts`, and `validation.ts` add the current-unit registry parser, active registry state, usage reads, decomposition, and validation helpers
- `engine.ts` prices token and request families through the registry-backed path while preserving the existing public result shape
- extractor output is validated and normalized as a plain usage object
- provider-array runtime updates remain supported and keep the generated registry active

Out of scope for this PR:
- Phase 3 wrapped remote payloads
- future text/image/video/cache-by-modality units
- public runtime unit mutation APIs
- validation trust markers/fingerprints, WeakMap caches, dirty sets, or decomposition caches
- Python behavior changes

## Testing

- `make package-data`
- `npm run typecheck --workspace=packages/js`
- `npm run test --workspace=packages/js`
- `npm run build --workspace=packages/js`

Final verification also confirmed:
- `packages/js/src/data.ts` exports provider data only
- `packages/js/src/dataUnits.ts` exports unit-family data only
- `prices/data.json` and `prices/data_slim.json` remain bare provider arrays


packages/js/src/units.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export function getUnit(usageKey: string): UnitDef {
+  for (const family of Object.values(activeFamilies)) {
+    const unit = family.units[usageKey]
+    if (unit) return unit
+  }
+  throw new Error(`Unknown unit usage key: ${usageKey}`)
+}
+
+export function getUnitForPriceKey(priceKey: string): UnitDef {
```
alexmojaki, line 89: this should be optimised

packages/js/src/units.ts
```diff
@@ -0,0 +1,153 @@
...
+export function getAllUsageKeys(): Set<string> {
+  return new Set(Object.values(activeFamilies).flatMap((family) => Object.keys(family.units)))
+}
+
+export function getReportedUsageKeys(): Set<string> {
+  return new Set([...getAllUsageKeys()].filter((usageKey) => usageKey !== 'requests'))
+}
+
+export function getUnit(usageKey: string): UnitDef {
+  for (const family of Object.values(activeFamilies)) {
```
alexmojaki: this should be optimised

packages/js/src/api.ts
```diff
@@ -24,13 +25,25 @@ function setProviderData(data: ProviderDataPayload) {
   }
   if ('then' in data) {
     providerDataPromise = data
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    data.then((data) => {
-      if (data !== null) {
+      .then((data) => {
```
alexmojaki, line 31: why has this area changed much? seems like adding validateExtractorDestinations should be simple

packages/js/src/units.ts
alexmojaki: (AI) Addressed in ef4d5c3 and finalized in 54c335b. These lookups now read directly from UnitRegistry indexes instead of scanning parsed family data.

packages/js/src/api.ts
alexmojaki: (AI) The extra change here is for async update semantics, not just the synchronous validation call. Runtime updates now validate extractor destinations before replacing providerData; null async updates keep the existing provider data; and validation failures reject waitForUpdate() once while leaving the current provider data active. I left it as-is because activation.test covers those rollback/use-time validation cases.

packages/js/src/units.ts
alexmojaki: (AI) Addressed in ef4d5c3 and finalized in 54c335b. The active registry now owns all usage/price key indexes, and the parsed-family compatibility path has been removed.

packages/js/src/__tests__/calcPrice.test.ts
```diff
@@ -54,11 +54,9 @@ describe('Core Price Calculation Function', () => {
       const inputPrice = uncachedInputPrice + cacheWritePrice + cacheReadPrice
       const outputPrice = (500 * 2.0) / 1_000_000
       const totalPrice = inputPrice + outputPrice
-      expect(result).toMatchObject({
-        input_price: inputPrice,
-        output_price: outputPrice,
-        total_price: totalPrice,
-      })
+      expect(result.input_price).toBeCloseTo(inputPrice)
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 3e8b4c7

PR #369, opened 2026-05-13
Implement Phase 4 registry polish hardening

alexmojaki:
## Summary
- add Phase 4 registry-derived provider YAML schema suggestions for price keys and extractor destinations
- support dynamic registered price-key kwargs on dataclass ModelPrice subclasses
- make CLI price labels, line formatting, split values, plain output, and table column discovery registry-driven

## Tests
- uv run pytest tests/test_units.py -k token
- uv run pytest tests/test_package_schema.py
- uv run pytest tests/test_units.py -k extractor
- uv run pytest tests/test_units.py -k "subclass or ModelPrice or model_price"
- uv run pytest tests/test_units.py -k "ModelPrice or model_price or dataclass"
- uv run pytest tests/test_cli.py
- uv run basedpyright

---
## Summary by cubic
Implements Phase 4 polish with a registry‑driven unit system across JS and Python: Python `ModelPrice` now stores price keys as plain attributes with runtime normalization of provider price maps, the provider YAML schema and CLI derive from the registry, runtime trusts compatible registry updates, JS tightens usage validation and guards async update races, and validation fails closed on missing legacy keys.

- **New Features**
  - Provider YAML schema is registry‑driven: adds dynamic `ModelPrice` properties for registry price keys and sets `extractor.dest` to an enum of reported usage keys; includes modality/cache variants.
  - CLI price display is registry‑driven: labels derive from unit dimensions and per values (K/MTok); plain “Model Prices” uses the same formatter; table columns follow effective model price keys (including dynamic extras) in registry order.
  - Python `ModelPrice` stores all price keys as direct attributes; `__getattr__` returns `None` for absent registered keys; provider parsing normalizes price mappings into `ModelPrice`; dynamic key iteration uses the registry. Removed runtime JSON schema customization for `ModelPrice`.
  - JavaScript usage validation rejects non‑finite or negative numbers for registered usage keys during normalization and reads, with clear errors.
  - Pricing validation fails closed when a legacy built‑in price key appears in model prices but is missing from the active registry.
  - Runtime treats published unit registries as compatible/trusted: installs fetched units and rolls back only if provider parsing fails; no registry‑history diff or snapshot pinning is enforced. JS guards an async update so a stale rejected promise can’t hide a newer in‑flight update.

- **Migration**
  - If you host custom price data, serve wrapped JSON `{ providers, units }`; run `make package-data` to regenerate generated unit files.
  - Update extractor mappings so `dest` is a valid registry usage key; invalid keys now fail validation.
  - Custom `ModelPrice` subclasses can keep custom fields and may accept registered price‑key kwargs; `ModelPrice.is_free()` was removed. Do not rely on `dataclasses.fields(ModelPrice)`; use the registry to enumerate price keys.

<sup>Written for commit e2b79bbaa33dba54ca0518602c42dd622f319366. Summary will update on new commits.</sup>


packages/js/src/data.ts
```diff
@@ -2852,6 +2927,66 @@ export const data: Provider[] = [
...
+              {
+                type: 'array-match',
+                field: 'modality',
+                match: {
+                  equals: 'TEXT',
+                },
+              },
+              'tokenCount',
+            ],
+            dest: 'output_text_tokens',
```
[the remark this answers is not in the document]
alexmojaki: really? we already treat tooluseprompt tokens as output in main.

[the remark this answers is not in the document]
alexmojaki: cubic, please interpret my questions literally, not rhetorically. "really?" meant "please verify and confirm", not "shut up". you were right about this point.

prices/src/prices/package_data.py
```diff
@@ -52,38 +100,116 @@ def package_python_data(data_path: Path):
...
+            field_name for field_name in model_price.__pydantic_fields__ if getattr(model_price, field_name) is not None
+        }
+        extra_keys = {field_name for field_name, value in (model_price.model_extra or {}).items() if value is not None}
+        return declared_keys | extra_keys
+
+    if dataclasses.is_dataclass(model_price):
+        declared_keys = {
+            field.name
+            for field in dataclasses.fields(model_price)
+            if field.name != '_extra_prices' and getattr(model_price, field.name) is not None
```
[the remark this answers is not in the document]
alexmojaki: what?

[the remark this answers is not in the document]
alexmojaki: no saving that was fine. look at the thread i'm saying "what?" in, you're mixing them up.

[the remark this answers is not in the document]
alexmojaki: > P2: Dataclass subclass fields that are not registered price keys are being validated as price keys, causing false Unknown price key failures during packaging.

what?

packages/python/genai_prices/update_prices.py
```diff
@@ -159,7 +164,21 @@ def _update_prices(self):
...
         r = httpx.get(self.url, timeout=self.request_timeout)
         r.raise_for_status()
-        return data_snapshot.DataSnapshot(data.providers_schema.validate_json(r.content), from_auto_update=True)
+        raw_payload = json.loads(r.content)
+        if not isinstance(raw_payload, dict) or 'units' not in raw_payload or 'providers' not in raw_payload:
+            raise ValueError('Expected fetched prices payload to contain units and providers')
+
+        candidate_registry = UnitRegistry(cast(dict[str, Any], raw_payload['units']))
+        previous_registry = _get_registry()
+        _set_registry(candidate_registry)
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 4092d54

packages/python/genai_prices/types.py
```diff
@@ -680,8 +680,28 @@ class CalcPrice(TypedDict):
...
+            object.__setattr__(instance, '_extra_prices', extra_prices)
+        return instance
+
+
 @dataclass
-class ModelPrice:
+class ModelPrice(metaclass=ModelPriceMeta):
     """Set of prices for using a model"""
 
     input_mtok: Decimal | TieredPrices | None = None
```
alexmojaki: do these fields still add anything other than IDE assistance? the expectation is that they shouldn't, that new and old units work the same in all contexts, and that removing these should be the same. dataclasses.fields and other introspection is allowed to break since it doesn't work for new fields anyway.

packages/js/src/api.ts
```diff
@@ -22,18 +26,49 @@ function setProviderData(data: ProviderDataPayload) {
...
-      }
-    })
+      .then((data) => {
+        if (data === null) {
+          return providerData
+        }
+        return activateProviderData(data)
+      })
+      .catch((error: unknown) => {
+        providerDataPromise = Promise.resolve(providerData)
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in e676321

tests/test_custom_prices.py
```diff
@@ -259,10 +259,13 @@ def test_custom_model_price_constructor_still_accepts_declared_custom_fields() -
...
+def test_custom_model_price_constructor_accepts_undeclared_dynamic_keys() -> None:
+    price = CustomModelPrice(
+        input_mtok=Decimal('1'),
+        cache_image_read_mtok=Decimal('0.5'),
+        sausage_price=Decimal('3'),
+    )
+
+    assert price.input_mtok == Decimal('1')
+    assert price.sausage_price == Decimal('3')
+    assert price._extra_prices == {'cache_image_read_mtok': Decimal('0.5')}
```
alexmojaki:
```suggestion
    assert price.cache_image_read_mtok == Decimal('0.5')
```

packages/js/src/usage.ts
```diff
@@ -0,0 +1,81 @@
...
+export type NormalizedUsage = Usage
+
+export function normalizeUsage(obj: unknown): NormalizedUsage {
+  if (!isPlainObject(obj)) return {}
+
+  const registry = getActiveRegistry()
+  const usage: NormalizedUsage = {}
+  for (const usageKey of registry.reportedUsageKeys) {
+    const value = obj[usageKey]
+    if (typeof value === 'number') {
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in d888672

prices/src/prices/package_data.py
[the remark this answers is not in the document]
alexmojaki: (AI) No action needed; Cubic confirmed this is not a packaging bug.

tests/test_custom_prices.py
alexmojaki: (AI) Addressed in c81d9ad

packages/python/genai_prices/types.py
alexmojaki: (AI) The legacy dataclass fields still provide the old constructor signature, type/docs surface, and a candidate list for old built-in price keys. They should not be the only pricing path, and new registered keys already work through `_extra_prices`. I tightened the concrete fail-closed issue in 5f09220 so a missing registry entry for an old built-in field is validated as an unknown price key instead of being skipped. Making the legacy fields fully removable/IDE-only would be a larger compatibility cleanup.

prices/src/prices/package_data.py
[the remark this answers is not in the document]
alexmojaki: (AI) Thanks, no action needed.

packages/python/genai_prices/types.py
alexmojaki: > old constructor signature, type/docs surface

i'm fine with losing these. i want old and new units to behave similarly. i don't want the constructor signature to be misleading and imply only accepting old units.

> a candidate list for old built-in price keys

what does this mean?

tests/test_custom_prices.py
alexmojaki: don't assert _extra_prices anywhere, it's an implementation detail

tests/test_units.py
```diff
@@ -500,6 +524,20 @@ def test_collect_effective_model_price_keys_includes_unregistered_extras_for_val
     assert _collect_effective_model_price_keys(price, registry) == {'hovercraft_mtok'}
 
 
+def test_collect_effective_model_price_keys_includes_builtin_fields_missing_from_registry() -> None:
```
alexmojaki: no such special casing is needed

tests/test_units.py
```diff
@@ -520,6 +558,61 @@ class CustomModelPrice(ModelPrice):
...
 
+def test_model_price_subclass_declared_custom_fields_stay_out_of_extra_prices() -> None:
+    @dataclass
+    class CustomModelPrice(ModelPrice):
+        sausage_price: Decimal | None = None
+
+    price = CustomModelPrice(input_mtok=Decimal('1'), sausage_price=Decimal('3'))
+
+    assert price.sausage_price == Decimal('3')
+    assert price._extra_prices == {}
```
alexmojaki: test public behaviour instead only

packages/python/genai_prices/types.py
alexmojaki: (AI) Addressed in 50d6f3e. Runtime `ModelPrice` no longer declares the legacy price keys as dataclass fields; old keys like `input_mtok` and registry-added keys now share the same constructor/storage/attribute path. The “candidate list” I mentioned was the old dataclass field list fallback, and that path is gone now.

tests/test_custom_prices.py
alexmojaki: (AI) Addressed in 50d6f3e. Tests now assert observable price-key behavior and no longer assert `price._extra_prices`.

tests/test_units.py
alexmojaki: (AI) Addressed in 72fc6c9. Removed the obsolete special-case registry mismatch test and kept the generic unregistered-candidate validation coverage.

tests/test_units.py
alexmojaki: (AI) Addressed in 50d6f3e. This test now checks public behavior via attribute access and effective price-key collection instead of inspecting `_extra_prices`.

packages/python/genai_prices/types.py
```diff
@@ -981,25 +940,54 @@ def _compute_registry_priced_counts(priced_units: Sequence[UnitDef], usage: Usag
...
         if value is not None and price_key not in yielded_price_keys:
             yield price_key
 
 
+def _iter_priced_registered_units(model_price: ModelPrice, registry: UnitRegistry) -> Iterator[UnitDef]:
+    units = (unit for unit in registry.units.values() if getattr(model_price, unit.price_key) is not None)
+    yield from sorted(units, key=_model_price_unit_sort_key)
+
+
+def _model_price_unit_sort_key(unit: UnitDef) -> tuple[int, int, str, int, int, tuple[tuple[str, str], ...], str]:
```
alexmojaki: what is this for?

alexmojaki: (AI) Addressed in bc87d5d. It was only preserving the old dataclass-field display order; I removed it and now registered price keys follow registry order.

alexmojaki:
@cubic-dev-ai review

packages/python/genai_prices/types.py
```diff
@@ -680,63 +680,44 @@ class CalcPrice(TypedDict):
...
-    """price in USD per thousand requests"""
+@dataclass
+class ModelPrice(metaclass=ModelPriceMeta):
+    """Set of prices for using a model"""
 
+    # The active registry is the price-key list; legacy keys and future keys use
+    # the same internal storage so dataclass introspection is not misleading.
     _extra_prices: dict[str, Decimal | TieredPrices | None] = dataclasses.field(
         default_factory=dict, repr=False, compare=False
     )
```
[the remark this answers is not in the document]
alexmojaki: intentional

PR #381, opened 2026-05-20
Improve Python test coverage

alexmojaki:
## Summary
- Strengthen Python coverage around usage extraction and price validation edge cases.
- Cover `ExtractedUsage.calc_price()` when response data has no model and the caller does not provide one.
- Make optional nested usage extractor paths best-effort: if an optional mapping path hits the wrong intermediate shape, it now returns `None` like optional `ArrayMatch` paths instead of raising.
- Add regression coverage for `ModelPrice.calc_price()` rejecting inconsistent inclusive token buckets, including audio/cache buckets and output audio counts.

## Tests
- `uv run coverage run -m pytest && uv run coverage report`
- `uv run ruff check packages/python/genai_prices/types.py tests/test_extract_usage.py tests/test_price_calc.py tests/test_cli.py`
- `uv run basedpyright`


tests/test_extract_usage.py
```diff
@@ -211,6 +231,32 @@ def test_extract_usage_error(response_data: Any, error: str):
...
+
+
+def test_usage_extractor_errors_when_nested_optional_path_has_wrong_type():
+    extractor = UsageExtractor(
+        root='usage',
+        mappings=[UsageExtractorMapping(path=['totals', 'input_tokens'], dest='input_tokens', required=False)],
+    )
+
+    with pytest.raises(ValueError, match='Expected `usage.totals` value to be a dict, got int'):
+        extractor.extract({'model': 'test-model', 'usage': {'totals': 1}})
```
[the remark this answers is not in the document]
alexmojaki: (AI) Noted, tracking in questions.md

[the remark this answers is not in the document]
alexmojaki: right, required=False shouldn't raise

packages/python/genai_prices/_cli.py
alexmojaki, line 1: i expect this file to essentially look like the old _cli_impl and for the diff to show a rename, is that possible?

packages/python/genai_prices/_cli.py
```diff
@@ -17,29 +26,715 @@ def _missing_cli_dependency_message(package: str) -> str:
...
+    from rich.columns import Columns
+    from rich.console import Console
+    from rich.markup import escape
+    from rich.table import Table
+    from rich.text import Text
+    from rich_argparse import RichHelpFormatter
+except ModuleNotFoundError as exc:
+    package = (exc.name or '').split('.')[0]
+    if package in _OPTIONAL_CLI_PACKAGES:
+        print(_missing_cli_dependency_message(package), file=sys.stderr)
```
alexmojaki: inline _missing_cli_dependency_message

tests/test_extract_usage.py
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 3572cb8

packages/python/genai_prices/_cli.py
alexmojaki: (AI) Addressed in 8ad5426

packages/python/genai_prices/_cli.py
alexmojaki: (AI) Not possible in the final PR diff without rewriting history because `_cli.py` already exists on `main`; GitHub cannot show `_cli_impl.py` as renamed onto an existing path. The consolidated `_cli.py` now otherwise mirrors the old `_cli_impl.py` implementation, with only the top-level optional dependency import guard added.

packages/python/genai_prices/types.py
```diff
@@ -455,10 +455,8 @@ def _extract_path(
             else:
                 return None
         else:
-            if not _is_mapping(data):
-                raise ValueError(
-                    f'Expected `{_dot_path(data_path, error_path)}` value to be a dict, got {_type_name(data)}'
-                )
+            if not _expect_mapping(data, required, data_path, error_path):
```
alexmojaki, line 458: make sure js matches

tests/test_price_calc.py
```diff
@@ -306,3 +306,37 @@ def test_output_audio_usage():
...
+            'Uncached text input tokens cannot be negative',
+        ),
+        (
+            ModelPrice(output_mtok=Decimal('1'), output_audio_mtok=Decimal('2')),
+            Usage(output_tokens=1, output_audio_tokens=2),
+            'output_audio_tokens cannot be greater than output_tokens',
+        ),
+    ],
+)
+def test_model_price_rejects_inconsistent_inclusive_token_buckets(model_price: ModelPrice, usage: Usage, error: str):
```
alexmojaki: this area is actually changing, drop these and add pragmas instead

packages/python/genai_prices/types.py
alexmojaki: (AI) Addressed in e2369b9

tests/test_price_calc.py
alexmojaki: (AI) Addressed in 3507b7a

PR #391, opened 2026-06-02
Update deepseek-v4-pro to new permanent pricing

MarkusWendorf:
DeepSeek had a promotion running that was 75% cheaper than baseline. However they decided to make the discount permanent (https://x.com/deepseek_ai/status/2057854261699195173).

This change reflects the pricing published here: https://api-docs.deepseek.com/quick_start/pricing/


alexmojaki:
Thanks!

PR #401, opened 2026-06-09
feat(openrouter): sync models from OpenRouter API

dmontagu:
Follow-up to #400 (which fixed `update_from_openrouter` crashing on adaptive `-1` pricing). Now that the pull runs, this syncs the catalog and curates the result.

## openrouter provider (auto-generated mirror)
- **+209 models** (587 → 796), including newer models that were missing: `z-ai/glm-4.7/5/5.1`, `minimax-m2/m2.1/m2.5/m2.7`, `kimi-k2.6`, `gpt-oss`, etc. — this answers the review question on whether these were curated in the openrouter file.
- Dropped **9 duplicates** the pull re-added for ids already covered by collapsed aliases (e.g. `gpt-4o-2024-08-06` under `gpt-4o`, `gemini-2.5-pro-preview` under `gemini-2.5-pro`) that otherwise conflicted with the alias match rules (build validation error).
- `collapse-models` merged 14 identical-price models into alias entries.

## native providers (curated, kept to each provider's own naming)
- Added **31 genuinely-new, correctly-named** models with real prices:
  - **deepseek**: v3.1-terminus, v3.2, v3.2-exp
  - **minimax**: minimax-01, minimax-m1, minimax-m3
  - **mistral**: codestral-2508, devstral-2512, ministral-3b/14b-2512, mistral-large-2512, mistral-small-2603/3.1/3.2, voxtral-small-24b-2507
  - **openai**: gpt-oss-20b/120b, gpt-oss-safeguard-20b, gpt-audio(-mini), gpt-chat-latest, gpt-5.4-image-2
  - **x_ai**: grok-4.20, grok-4.20-multi-agent, grok-build-0.1
  - **azure**: phi-4-mini-instruct; **google**: gemma-2-27b-it, gemma-4-26b/31b-it; **moonshotai**: kimi-k2; **perplexity**: sonar-pro-search
- **Skipped** OpenRouter-naming duplicates of existing curated models (`minimax-m2-her` vs `M2-her`, `kimi-k2-0905` vs `kimi-k2-0905-preview`, the `nova-*-v1` set vs `amazon.nova-*-v1:0`), Bedrock-incompatible AWS ids, and `:free`/zero-price entries (`gpt-oss:free`, `gemma:free`, `lyria-3` music models).
- Reverted the pull's redundant match-clause / metadata edits to existing models, so only new entries are added.

## Notes
- **No existing model prices changed.**
- Fixed an OpenRouter-sourced typo ("adherance" → "adherence").
- Updated the deepseek CLI list snapshot and `tests/dataset/usages.json` (`gpt-oss-120b` now resolves to a price).
- All green locally: build, collapse, no price discrepancies, Python tests (97.38% cov), JS (619), full pre-commit.


dsfaccini:
David's AICA here: reviewing from a Pydantic AI cost-calculation perspective, I found one blocker and one follow-up concern.

1. **OpenRouter model IDs are stripped, so Pydantic AI cannot price many of these new entries.**

   The OpenRouter provider data now contains entries such as:

   - `claude-fable-latest` instead of `~anthropic/claude-fable-latest`
   - `glm-4.7` instead of `z-ai/glm-4.7`
   - `gpt-latest` instead of `~openai/gpt-latest`
   - `minimax-m2.7` instead of `minimax/minimax-m2.7`

   Pydantic AI stores OpenRouter responses using the exact `response.model` value and then calls `genai_prices.calc_price(..., provider_api_url='https://openrouter.ai/api/v1')`. Existing Pydantic AI OpenRouter cassettes show response models like `z-ai/glm-4.6`, `openai/o3`, and `anthropic/...`, not stripped IDs.

   I confirmed this locally on this PR branch:

   ```py
   from genai_prices import Usage, calc_price

   usage = Usage(input_tokens=1000, output_tokens=1000)
   calc_price(usage, 'glm-5.1', provider_api_url='https://openrouter.ai/api/v1')      # hits
   calc_price(usage, 'z-ai/glm-5.1', provider_api_url='https://openrouter.ai/api/v1') # LookupError
   ```

   Root cause appears to be `prices/src/prices/source_openrouter.py::OpenRouterModel.model_id`, which strips the provider prefix with `self.id.split('/', 1)[1]`. That may be appropriate for native-provider curation, but the `openrouter` provider mirror needs to preserve `or_model.id` so it matches OpenRouter/Pydantic AI model names.

   **Verdict: net-positive -> do**: preserve full OpenRouter IDs in `prices/providers/openrouter.yml`; only use provider-native stripped IDs for native provider additions.

2. **The #400 negative-price predicate is still broader than the represented price fields.**

   `OpenRouterPricing.model_price()` only consumes `prompt`, `completion`, `input_cache_write`, and `input_cache_read`, but `has_negative_price()` skips on `audio`, `request`, `image`, `web_search`, and `internal_reasoning` too. If OpenRouter ever reports a dynamic sentinel for one of those unrepresented fields while token prices are valid, this would remove normal text-token cost support for that model.

   **Verdict: net-positive -> do**: narrow the predicate to the fields `model_price()` actually stores, or explicitly document that any dynamic/unrepresented cost component makes total cost intentionally unavailable.

I did not run the full genai-prices suite; this was source/diff review plus targeted `uv run python` checks against the generated data and Pydantic AI's OpenRouter response handling.

PR #422, opened 2026-06-24
fix: price compact GPT-5.2 date refs

### LINKED ISSUE #333: [LiteLLM/OpenRouter] pricing lookup fails for compact dated model refs
We’re seeing a pricing lookup mismatch when using `genai-prices` through the LiteLLM path.

In our case, the traced response model from OpenRouter is:

```json
"gen_ai.response.model": "openai/gpt-5.2-20251211"
```

PydanticAI eventually calls `genai_prices.calc_price(...)` with:

- `model_ref="openai/gpt-5.2-20251211"`
- `provider_id="litellm"`

From reading `data_snapshot.py`, the LiteLLM path strips the provider prefix and then looks up:

- `gpt-5.2-20251211`

The bundled aliases in `data.py` appear to be hardcoded and currently include entries such as:

- `gpt-5.2`
- `gpt-5.2-2025-12-11`

but not the compact-date form:

- `gpt-5.2-20251211`

So the lookup fails unless we normalize the model ref ourselves first.

It seems like one of these should happen in `genai-prices`:
- add compact-date aliases like `gpt-5.2-20251211`
- or normalize/strip compact date formats into the existing canonical form like `gpt-5.2-2025-12-11` or `gpt-5.2`

**Minimal example**

```python
from genai_prices import calc_price
from genai_prices.types import Usage

usage = Usage(input_tokens=1000, output_tokens=100)

calc_price(
    usage,
    "openai/gpt-5.2-20251211",
    provider_id="litellm",
)
```

**Expected**

This should resolve to the same pricing as `gpt-5.2-2025-12-11`, or the compact-date form should be normalized automatically before lookup.

**Actual**

Lookup fails for the compact-date model ref.

**Versions**

- Python: `3.13.12`
- Pydantic AI: reproduced on `1.70.0` and `1.23.0`
- LiteLLM: `1.80.0`

fengjikui:
## Summary

- Fixes #333.
- Adds compact-date GPT-5.2 aliases for the LiteLLM provider path and direct OpenRouter matching.
- Regenerates the Python and JavaScript package data from the provider YAML files.
- Covers every added alias with model and exact-price assertions.

## Root cause

LiteLLM identifies the provider from a model ref such as `openai/gpt-5.2-20251211`, strips the `openai/` prefix, and asks the OpenAI provider to match `gpt-5.2-20251211`. The bundled data only recognized the hyphenated date form (`gpt-5.2-2025-12-11`), so real OpenRouter trace model refs raised `LookupError`.

This branch was rebuilt on `main@cc40e93`. The reported LiteLLM path and the equivalent OpenRouter API URL path both failed on that baseline and resolve to the existing GPT-5.2 pricing after this change.

## Validation

- Targeted alias and price tests: `6 passed`
- `env -u ALL_PROXY -u HTTPS_PROXY -u HTTP_PROXY -u NO_PROXY -u http_proxy -u https_proxy -u no_proxy make all` (`530 passed`, `41 xfailed`, `100.00%` coverage; Ruff and basedpyright passed)
- `npm run ci` (`630 passed`, `1 skipped`)
- `env SKIP=no-commit-to-branch uvx pre-commit run --color=never --all-files --verbose`
- `git diff --check`


dsfaccini:

Superseded by the merged consolidation in #616 (@dsfaccini). The OpenRouter alias and pricing regression work were incorporated, with @fengjikui preserved as a parsed Git co-author and #422 (@fengjikui) credited explicitly. Thank you for the contribution!

PR #434, opened 2026-07-01
Auto-release on data/source changes to main

dsfaccini:
## What

Auto-cut a patch release whenever **packaged data or package source** lands on `main`, so the Slack ask ("do releases whenever we merge to main") is satisfied without anyone hand-bumping versions or pushing tags.

The goal isn't to push people to `@latest` — it's the opposite: keep a **fresh, pinnable version always available** so a consumer can bump `genai-prices` to a specific version in the same PR where they switch a model, instead of either floating `@latest` or hitting the `raw.githubusercontent` `data.json` from prod (which not everyone wants to do).

## How

New `.github/workflows/release.yml`, on `push: main`:

1. Triggers only on changes to `packages/python/genai_prices/**` or `packages/js/src/**` (the packaged artifacts — `data.py` / `data.ts` — and source).
2. `uv version --package genai-prices --bump patch` + `npm version patch --workspace=packages/js` → bumps all 4 version files (`pyproject.toml`, `package.json`, `uv.lock`, `package-lock.json`) in lockstep.
3. Commits `Release vX.Y.Z` + tags + pushes to `main`.
4. The **tag push** triggers the existing `release-pypi` / `release-npm` jobs in `ci.yml` that actually publish. No change to the publish path.

**No re-trigger loop:** the version files the bump touches all sit *outside* the trigger `paths`, so a release commit can't re-fire the workflow. `concurrency: release` serialises back-to-back merges so they don't race on a version number.

## ⚠️ Required one-time org setup (workflow is inert until done)

The bump commit + tag push to **protected `main`**, so the default `GITHUB_TOKEN` won't do (can't bypass the ruleset; its pushes also wouldn't re-trigger CI). Need:

- A **GitHub App** with `contents: write`, installed on this repo.
- Add that App to the **`main` ruleset bypass list**.
- Repo **variable** `RELEASE_APP_ID` + repo **secret** `RELEASE_APP_PRIVATE_KEY`.

(A fine-grained PAT with the same bypass works as a simpler alternative — swap the `create-github-app-token` step for `token: ${{ secrets.RELEASE_PAT }}` on checkout.)

Opening as **draft** until the above is configured.

## Notes / deliberate scope

- **Trigger = packaged-data/source paths, not every merge.** Docs/CI/test-only merges don't cut a release. Covers both "latest prices" and "latest patches".
- **One redundant CI run per release:** the bump commit's push to `main` re-runs `ci.yml` tests (release jobs skipped — not a tag). Harmless; left as-is to avoid a `[skip ci]` that would also skip the tag's publish run (same SHA).
- **`@latest` is still not advertised anywhere** — this PR doesn't touch docs to encourage floating; it just makes pinning painless. Happy to add a short "two channels" note (pin a release vs. on-demand data fetch) in a follow-up if wanted.
- **Locally verified:** both bump commands touch exactly the 4 version files and land on matching `0.0.68`; `zizmor` passes on the workflow.

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/434?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,85 @@
+# Auto-cuts a patch release whenever packaged data or package source lands on main,
+# so consumers can pin a fresh, specific version (instead of floating @latest or
+# hitting the raw.githubusercontent data.json from prod).
+#
+# Flow: bump patch across both packages -> commit + tag on main -> the tag push
+# triggers the `release-pypi` / `release-npm` jobs in ci.yml that actually publish.
+#
+# Required org setup (one-time):
+#   - A GitHub App with `contents: write`, installed on this repo.
+#   - Add the App to the `main` ruleset's bypass list (it pushes the bump commit + tag
+#     directly to the protected branch).
+#   - Repo variable  RELEASE_APP_ID            = the App's id
+#   - Repo secret     RELEASE_APP_PRIVATE_KEY  = the App's private key
+# (A fine-grained PAT with the same bypass works too; swap the app-token step for a
+#  `token: ${{ secrets.RELEASE_PAT }}` on checkout.)
+
+name: Release
+
+on:
+  push:
+    branches: [main]
+    # Only the packaged artifacts / source — NOT the version files the bump itself
+    # touches (pyproject.toml, package.json, *.lock), so a release commit can't
+    # re-trigger this workflow.
+    paths:
+      - "packages/python/genai_prices/**"
+      - "packages/js/src/**"
+
+permissions:
+  contents: read
+
+# Serialise releases so two quick merges don't race on the same version number.
+concurrency:
+  group: release
+  cancel-in-progress: false
+
+jobs:
+  tag:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        id: app-token
+        with:
+          app-id: ${{ vars.RELEASE_APP_ID }}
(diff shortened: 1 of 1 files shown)


dsfaccini:
David's AICA here: flagging the **required one-time setup** so this doesn't silently no-op if merged. Keeping the PR in **draft** until these are done.

**This workflow is inert until all of the below exist** — the bump commit + tag push to protected `main`, which the default `GITHUB_TOKEN` cannot do (can't bypass the ruleset, and its pushes wouldn't re-trigger the publish CI):

- [ ] A **GitHub App** with `contents: write`, installed on this repo
- [ ] That App **added to the `main` ruleset's bypass list** (it pushes the release commit + tag directly to the protected branch)
- [ ] Repo **variable** `RELEASE_APP_ID` = the App's id
- [ ] Repo **secret** `RELEASE_APP_PRIVATE_KEY` = the App's private key

Alternative to the App: a fine-grained **PAT** with the same `contents: write` + ruleset bypass, stored as `RELEASE_PAT`, swapped in on the checkout step (one-line change — say the word and I'll adjust the YAML).

Until then, merging this changes nothing (no permissions to push), so it's safe to sit as a draft while the org config is sorted.

Kludex:
Closing because the release design is superseded by the tag-derived release flow merged in [#622](https://github.com/pydantic/genai-prices/pull/622).

PR #447, opened 2026-07-02
Introduce when/values conditional prices with per-unit first-match

Kludex:
## Summary

Implements the Conditional Prices spec: replaces the whole-set, last-match-wins `ConditionalPrice` mechanism with `when` + `values` entries evaluated **per-unit, first-match**. An entry lists only the prices that differ under a condition; every other unit falls through to a later entry, eliminating the duplication of repeating a full price block per condition.

Stacked on Alex's unit-registry Phase 4 branch (#369). Provider-Level Prices is a separate follow-up.

## What changed

- **Data model**: `ConditionalPrice` gains an optional `when` clause and renames `prices` → `values`. For each unit, the first eligible entry (top to bottom) that defines it wins; unlisted units are zero-priced.
- **Conditions**: literal-value equality shorthand plus `eq` / `gte` / `lte` / `gt` / `lt` / `in` operators (combine `gte`/`lte` for a range). Fixed parameter set: `service_tier`, `batch`, `cache_ttl`, `fast_mode`, `inference_geo`, `region`.
- **Request-level `price_context`**: threaded through `calc_price`, `ModelInfo.calc_price`, `ExtractedUsage.calc_price`, and `DataSnapshot.calc` (additive, backward compatible) and echoed on `PriceCalculation.price_context`.
- **Unchanged mechanisms**: `StartDateConstraint`, `TimeOfDateConstraint`, and `TieredPrices` keep their current behavior. An entry is eligible when its date/time constraint is active (or absent) **and** its `when` matches.
- **JS parity**: `getActiveModelPrice` in the JS engine does the same per-unit first-match with matching operators and a `priceContext` argument.
- **Data + schema**: provider YAML migrated (date/time conditional entries reordered so the unconditional entry is last, which preserves current pricing under first-match), and `data.json` / `data_slim.json` / `data.py` / `data.ts` / schemas regenerated.

## Compatibility

- Public API changes are additive; existing callers that pass no `price_context` are unaffected.
- Consumers that read only `total_price` / `extract_usage` / `find_provider().models` (pydantic-ai, the gateway, logfire) are unaffected at the API level.
- Note for release coordination: the conditional entry shape in `data.json` changed (`prices` → `values`), so an older pinned engine reading a newer `data.json` would misprice the 5 conditional models. A `genai-prices` version bump should ship alongside the data change for runtime `data.json` consumers.

## Test Plan

- `uv run pytest` — 651 passed (incl. new `tests/test_conditional_prices.py`)
- `uv run basedpyright` — clean
- `npm test` (packages/js) — 718 passed (incl. new `conditionalPrices.test.ts`)
- `npm run typecheck` / `npm run lint` (packages/js) — clean

## AI Disclaimer

This PR was developed with the assistance of either Claude or Codex. I've reviewed and verified the changes.

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/447?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kludex:
Closing this obsolete conditional-pricing implementation as stale. The current implementation is [#547](https://github.com/pydantic/genai-prices/pull/547).

PR #448, opened 2026-07-02
Add provider-level prices inherited by models

Kludex:
## Summary

Implements the Provider-Level Prices spec, the companion to the `when`/`values` conditional-prices work. Adds an optional `prices` key on the provider (sibling to `models`), inherited by every model in that provider.

Stacked on #447 (conditional prices). Review that PR first for the shared `when`/`values` machinery this builds on.

## What changed

- **Provider-level prices**: a new optional `prices` field on `Provider`, same shape as model prices (`ModelPrice | list[ProviderConditionalPrice]`). All models inherit it.
- **Merge rule**: model-level prices win. The engine resolves model-level prices, then fills in any unit the model does not price from the provider-level prices.
- **Model matching**: provider-level `when` clauses additionally support a `model` parameter matched against the model id using the existing `MatchLogic` operators (`starts_with`, `or`, `in`, `eq`, ...), so a provider can price a unit differently by model category without repeating it on every model. The `model` parameter is only meaningful at the provider level.
- **JS parity**: `getActiveModelPrice` takes the provider and performs the same collection + merge + model matching.
- **Schema**: adds `ProviderConditionalPrice` (its `when` allows `model`) and `Provider.prices`, with a validator requiring exactly one unconditional entry, last.

No provider YAML data is added in this PR: the shared registry does not yet define tool-call units (`web_search`, etc.), and there is no current data gap that needs provider-level prices, so the feature is exercised via tests rather than churning provider files. Generated `data.json` / `data.py` / `data.ts` are unchanged.

## Compatibility

Purely additive. Providers without a `prices` key behave exactly as before; the generated data is unchanged.

## Test Plan

- `uv run pytest` — 658 passed (incl. new `tests/test_provider_prices.py`: inheritance, model-override, provider-level model matching, raw-JSON parsing, build-schema validation)
- `uv run basedpyright` — clean
- `npm test` (packages/js) — 722 passed (incl. new `providerPrices.test.ts`)
- `npm run typecheck` / `npm run lint` (packages/js) — clean

## AI Disclaimer

This PR was developed with the assistance of either Claude or Codex. I've reviewed and verified the changes.

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/448?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kludex:
Closing as stale because this branch depends on the obsolete [#447](https://github.com/pydantic/genai-prices/pull/447) stack and its description identifies no current data gap. Provider-level prices can be revisited with a concrete use case.

PR #326, opened 2026-03-25
Plan: Cohere extractor billed_units vs tokens (#324)

adtyavrdhn:
Research plan for #324 Gap #2. Cohere extracts from `billed_units` (correct for cost calc) but pydantic-ai uses `tokens`. Proposes adding a `tokens` api_flavor. See PLAN.md for full analysis.

Generated with [Claude Code](https://claude.com/claude-code)

Issue #324, opened 2026-03-25
Usage extractor gaps and semantics issues affecting pydantic-ai migration to RequestUsage.extract()


adtyavrdhn:
## Context

[pydantic/pydantic-ai#4818](https://github.com/pydantic/pydantic-ai/issues/4818) is migrating remaining models to use `RequestUsage.extract()` via genai-prices.

This issue needs to stay scoped to actual `genai-prices` blockers and correctness issues:

- `groq` is no longer a remaining migration target there; [PR #4945](https://github.com/pydantic/pydantic-ai/pull/4945) merged on April 7, 2026
- `google` / Gemini direct already uses `RequestUsage.extract()` in `pydantic-ai`; the Google item below is a separate correctness fix, not a remaining migration blocker
- The remaining manual migrations that may still depend on `genai-prices` work are `bedrock`, `cohere`, `huggingface`, and likely `mistral`

### 1. AWS Bedrock: default extractor missing cache token mappings

The default extractor only maps `inputTokens` and `outputTokens`, but Bedrock still returns `cacheReadInputTokens` and `cacheWriteInputTokens` in `TokenUsage`.

- AWS docs: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_TokenUsage.html
- Current extractor: `prices/providers/aws.yml`

This is a real gap, but the fix is **not** just adding `cache_read_tokens` / `cache_write_tokens`.

`genai-prices` pricing logic assumes `input_tokens` includes all input-side tokens and then subtracts cache tokens to derive uncached input. So for Bedrock, the default extractor likely needs Anthropic-style semantics:

- `inputTokens -> input_tokens`
- `cacheReadInputTokens -> cache_read_tokens`
- `cacheWriteInputTokens -> cache_write_tokens`
- and cache read/write tokens also summed into `input_tokens`

Without that, pricing can become incorrect for models with separate cache pricing.

### 2. HuggingFace: generated providers have no `default` flavor

All generated Hugging Face provider YAMLs only expose `api_flavor: chat`.

But `UsageExtractor.api_flavor` defaults to `'default'`, and callers using `extract_usage(..., api_flavor='default')` will fail flavor lookup.

Current generator behavior:
- `prices/src/prices/source_huggingface.py` copies only the OpenAI `chat` extractor

Fix direction:
- either generate a duplicate `default` flavor
- or make the generated extractor itself be `default`

### 3. HuggingFace: possible provider-resolution gap even after the flavor fix

[...]

PR #453, opened 2026-07-07
Add Cohere `tokens` api_flavor for raw token extraction

adtyavrdhn:
Part of #324 (gap 4, the Cohere semantics decision). Unblocks the Cohere migration in pydantic/pydantic-ai#4818.

Decision implemented, per the analysis in #326: the `default` flavor stays on `usage.billed_units` so pricing remains correct (billed units are what Cohere charges for; raw tokens can be ~40x larger). A new opt-in `api_flavor: tokens` extracts raw counts for callers that report usage rather than price it: `usage.tokens.input_tokens`, `usage.tokens.output_tokens`, and top-level `usage.cached_tokens` into `cache_read_tokens`. All mappings are `required: false` because Cohere omits the `tokens` sub-object in some responses.

pydantic-ai will call this with `api_flavor='tokens'` to keep its current raw-token reporting unchanged, with billed units preserved in its details dict.

Tests cover both flavors on the same body, cached tokens without the `tokens` sub-object, the no-usage error path, default-flavor pricing, and the engine's behavior on optional float values. Recorded corpus fixtures confirm the raw vs billed divergence (2190 vs 431 input tokens on one response).

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/453?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


adtyavrdhn:
Addressed the line anchor in 76fa981 (usage.py#L14).

On the float concern: the Python engine's int-only extraction is pre-existing and deliberate, and this PR pins it with a unit test (test_usage_extractor_skips_optional_float_value). Real Cohere wire JSON serializes these counts as integers (verified against the recorded corpus in tests/dataset). The float typing only appears when callers dump SDK objects; the pydantic-ai migration that consumes this flavor normalizes those to int before extraction. Widening the engine to accept integral floats would be an engine-wide behavior change out of scope here (and would also diverge further from the JS engine, which already accepts any number).

PR #465, opened 2026-07-15
Share model metadata across provider offerings

adtyavrdhn:
## Summary

Adds an authoring-only `canonical_model` link from a provider offering to its underlying model record, and fills a missing `context_window` from it at build time. An explicit value on the offering always wins, and the link is not published.

pydantic-ai gets effective context windows through the existing lookup (pydantic/pydantic-ai#4611), with no second lookup or name translation:

```python
provider, model = snapshot.find_provider_model(model_name, None, model_system, None)
context_window = model.context_window
```

## Changes

- `canonical_model` must resolve to an exact `provider/model` record. Unknown and chained references fail the build. Links resolve before removed models are filtered, so an inactive native record can still supply metadata to an active hosted offering.
- 51 relationships: AWS→Anthropic (31), Google→Anthropic (11), Azure→OpenAI (8), Fireworks→DeepSeek (1). A link is only set where the value holds for every model the record matches; ambiguous rows are left out, e.g. OpenRouter's DeepSeek V4 Pro, whose selectable backends range from 512K to 1M.
- Splits stable `o1` (200K) from `o1-preview` (128K) and adds documented limits to `o1`, `o3`, `o3-mini`, and `o4-mini`, which Azure inherits.
- Explicit overrides where a host documents a different limit (e.g. Bedrock Sonnet 4.5 at 200K), each with a first-party link in the YAML.
- Only `context_window` is inherited for now. Other fields can be added deliberately later without changing the consumer lookup.

Refs #260 and pydantic/pydantic-ai#4611.


packages/python/genai_prices/_model_metadata.json
```diff
@@ -0,0 +1,259 @@
+{
+  "version": 1,
+  "source": "https://models.dev/models.json",
+  "source_sha256": "b3bc0c06926df76a76ec498f5c38abbd490b68090e6dc7d5c017c00b0534ec0d",
```
alexmojaki, line 4: do these fields have a purpose

packages/python/genai_prices/_model_metadata.json
```diff
@@ -0,0 +1,259 @@
...
+    "anthropic/claude-fable-5": [1000000, null, 128000],
+    "anthropic/claude-haiku-4-5": [200000, null, 64000],
+    "anthropic/claude-haiku-4-5-20251001": [200000, null, 64000],
+    "anthropic/claude-opus-4-0": [200000, null, 32000],
+    "anthropic/claude-opus-4-1": [200000, null, 32000],
+    "anthropic/claude-opus-4-1-20250805": [200000, null, 32000],
+    "anthropic/claude-opus-4-20250514": [200000, null, 32000],
+    "anthropic/claude-opus-4-5": [200000, null, 64000],
+    "anthropic/claude-opus-4-5-20251101": [200000, null, 64000],
+    "anthropic/claude-opus-4-6": [1000000, null, 128000],
```
alexmojaki, line 51: this doesn't feel very provider-independent, you have to provide an exact provider string. how does it help?

packages/python/genai_prices/__init__.py
```diff
@@ -5,11 +5,21 @@
...
 from .types import Usage
 from .update_prices import UpdatePrices, wait_prices_updated_async, wait_prices_updated_sync
 
 __version__ = _metadata_version('genai_prices')
-__all__ = 'Usage', 'calc_price', 'UpdatePrices', 'wait_prices_updated_sync', 'wait_prices_updated_async', '__version__'
+__all__ = (
+    'Usage',
+    'ModelMetadata',
+    'calc_price',
+    'get_model_metadata',
```
alexmojaki, line 17: maybe let's not expose these from the top level yet, let it be a bit experimental at first

packages/python/genai_prices/model_metadata.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    model_ref = model_ref.strip().lower()
+    if not model_ref:
+        return None
+    return _get_bundled_model_metadata().get(model_ref)
+
+
+@cache
+def _get_bundled_model_metadata() -> dict[str, ModelMetadata]:
+    data = files(__package__).joinpath('_model_metadata.json').read_bytes()
```
alexmojaki, line 53: please double check that the json file is included in the built python package/wheel

packages/python/README.md
```diff
@@ -87,6 +87,23 @@ price = extracted_usage.calc_price()
 print(price.total_price)
 ```
 
+### `get_model_metadata`
```
alexmojaki, line 90: i think leave this out of docs while we experiment with it

packages/python/README.md
```diff
@@ -87,6 +87,23 @@ price = extracted_usage.calc_price()
...
+from genai_prices import get_model_metadata
+
+metadata = get_model_metadata('anthropic/claude-sonnet-4-5')
+assert metadata is not None
+assert metadata.context_window == 200_000
+assert metadata.max_output_tokens == 64_000
+```
+
+The lookup is case-insensitive and returns `None` when token-limit metadata is unavailable. It is backed by a
+single bundled snapshot, so normal imports and lookups do not make network requests.
```
alexmojaki, line 105: why are network requests being mentioned twice as if the rest of the package makes requests by default

tests/test_lifecycle.py
```diff
@@ -62,3 +63,10 @@ def test_remote_payloads_remain_provider_arrays():
         assert isinstance(payload, list)
         assert payload
         assert all(isinstance(provider, dict) for provider in payload)
+
+
+def test_model_metadata_has_single_packaged_artifact():
```
alexmojaki, line 68: this test is just noise

.gitattributes
```diff
@@ -0,0 +1 @@
+packages/python/genai_prices/_model_metadata.json linguist-generated=true
```
alexmojaki, line 1: what does this mean?

packages/python/genai_prices/_model_metadata.json
adtyavrdhn: Agreed. The consumer now passes its raw model reference plus provider to the existing matcher. `canonical_model` is only an internal authoring link used at build time, so Pydantic AI does not provide an exact canonical/provider string or perform a second lookup.

packages/python/genai_prices/_model_metadata.json
adtyavrdhn: Removed the standalone snapshot entirely, including these source fields. The new design reuses provider data during the existing build.

packages/python/genai_prices/model_metadata.py
adtyavrdhn: The separate JSON artifact has been removed. Effective context windows are materialized into the existing generated package data; I also rebuilt and inspected the wheel.

packages/python/genai_prices/__init__.py
adtyavrdhn: Removed. This no longer adds any top-level or experimental runtime API.

packages/python/README.md
adtyavrdhn: Removed the documentation changes along with the experimental public API.

packages/python/README.md
adtyavrdhn: Removed. There is no new network-backed importer or related documentation in the revised design.

.gitattributes
adtyavrdhn: Removed the `.gitattributes` change; the generated snapshot it was intended for no longer exists.

tests/test_lifecycle.py
adtyavrdhn: Removed the file-existence lifecycle test. The replacement tests cover build behavior and the actual `DataSnapshot.find_provider_model()` consumer path.

prices/src/prices/prices_types.py
```diff
@@ -178,6 +178,8 @@ class ModelInfo(_Model):
     """Description of the model"""
     match: MatchLogic
     """Boolean logic for matching this model to any identifier which could be used to reference the model in API requests"""
+    canonical_model: IdField | None = Field(default=None, exclude=True)
```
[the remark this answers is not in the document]
adtyavrdhn: Valid catch. `canonical_model` now uses a dedicated 201-character reference type, and a boundary test covers 100-character provider and model IDs.

prices/src/prices/build.py
```diff
@@ -69,6 +70,24 @@ def build():
...
+            if canonical_ref := model.canonical_model:
+                canonical = models.get(canonical_ref)
+                if canonical is None:
+                    raise ValueError(
+                        f'Model `{provider.id}/{model.id}` references unknown canonical model `{canonical_ref}`'
+                    )
+                if canonical.canonical_model is not None:
+                    raise ValueError(f'Canonical model `{canonical_ref}` must not reference another canonical model')
+                if model.context_window is None and canonical.context_window is not None:
+                    model.context_window = canonical.context_window
```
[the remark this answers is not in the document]
adtyavrdhn: This is intentional. `canonical_model` represents a verified same-underlying-model relationship, not a promise that the canonical record currently contains a context window. Requiring that one field would prevent us from recording identity now and sharing other or future model metadata later. Missing/stale references are still rejected because the canonical `provider/model` must resolve exactly; when metadata exists, the provider offering inherits it, and an explicit provider value wins.

prices/providers/fireworks.yml
```diff
@@ -86,6 +86,7 @@ models:
       output_mtok: 0.28
 
   - id: deepseek-v4-pro
+    canonical_model: deepseek/deepseek-v4-pro
```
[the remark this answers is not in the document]
adtyavrdhn: Confirmed against the official Fireworks model catalog: this offering has a 1,048,576-token context window (not 1,040,000). Added that exact explicit provider override plus a consumer-shaped regression assertion in ad0cc3c. The canonical identity remains, while the Fireworks-specific value correctly wins. Source: https://fireworks.ai/models/fireworks/deepseek-v4-pro

prices/providers/azure.yml
```diff
@@ -152,6 +154,7 @@ models:
       output_mtok: 4.4
 
   - id: o3-2025-04-16
+    canonical_model: openai/o3
```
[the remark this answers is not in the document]
adtyavrdhn: These links are intentionally useful independently of whether the canonical records contain context metadata today. `canonical_model` represents verified same-underlying-model identity; context inheritance is only its first use. Requiring every link to immediately materialize context would couple identity to one optional field and prevent later metadata sharing. Exact targets are validated, and explicit provider metadata wins.

.github/workflows/ci.yml
```diff
@@ -163,10 +163,29 @@ jobs:
...
           for schema in prices/new_data/v2/data.schema.json prices/new_data/v2/data_slim.schema.json; do
-            if ! git diff --quiet "${BASE_SHA}" -- "$schema"; then
+            git show "${BASE_SHA}:$schema" > /tmp/base-schema.json
+            if ! python3 - "$schema" /tmp/base-schema.json <<'PY'
+          import json
+          import sys
+
+          def strip_descriptions(node):
+              if isinstance(node, dict):
+                  return {k: strip_descriptions(v) for k, v in node.items() if k != 'description'}
```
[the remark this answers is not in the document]
adtyavrdhn: Fixed in 76adb13 and 1a159bb. The comparator preserves property-name keys and strips descriptions only in schema positions. Tests cover a property literally named description plus object-valued const, enum, default, and examples.

.github/workflows/ci.yml
```diff
@@ -163,10 +163,29 @@ jobs:
...
+              if isinstance(node, list):
+                  return [strip_descriptions(item) for item in node]
```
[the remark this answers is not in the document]
adtyavrdhn: Fixed in 76adb13 and 1a159bb. The comparator is now schema-context-aware, preserves description properties, and has negative contract-change tests.

prices/new_data/v2/data.json
```diff
@@ -1 +1 @@
...eek V3.2","match":{"equals":"deepseek/deepseek-v3.2"},"prices":{"input_mtok":0.2288,"output_mtok":0.3432}},{"id":"deepseek/deepseek-v3.2-exp","name":"DeepSeek V3.2 Exp","match":{"equals":"deepseek/deepseek-v3.2-exp"},"prices":{"input_mtok":0.27,"output_mtok":0.41}},{"id":"deepseek/deepseek-v4-flash","name":"DeepSeek V4 Flash","match":{"equals":"deepseek/deepseek-v4-flash"},"prices":{"input_mtok":0.0983,"cache_read_mtok":0.0197,"output_mtok":0.1966}},{"id":"deepseek/deepseek-v4-pro","name":"DeepSeek V4 Pro","match":{"equals":"deepseek/deepseek-v4-pro"},"prices":{"input_mtok":0.435,"cache_read_mtok":0.003625,"output_mtok":0.87}},{"id":"devstral-small","name":"Devstral Small","match":{"equals":"devstral-small"},"prices":{"input_mtok":0.06,"output_mtok":0.12}},{"id":"devstral-small:free","name":"Devstral Small (free)","match":{"equals":"devstral-small:free"},"prices":{}},{"id":"dobby-mini-unhinged-plus-llama-3.1-8b","name":"Dobby Mini Plus Llama 3.1 8B","match":{"equals":"dobby-mini-unhinged-plus-llama-3.1-8b"},"prices":{"input_mtok":0.2,"output_mtok":0.2}},{"id":"dolphin-mixtral-8x22b","name":"Dolphin 2.9.2 Mixtral 8x22B 🐬","match":{"equals":"dolphin-mixtral-8x22b"},"prices":{"input_m
```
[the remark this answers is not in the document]
adtyavrdhn: Fixed in 76adb13, ab263b3, and 1a159bb. Google Sonnet 4, 4.5, and 4.6 now have disjoint direct matchers, correct provider context windows, preserved aliases, tiered price coverage, and Python/JS regressions.

.github/workflows/ci.yml
```diff
@@ -163,10 +163,29 @@ jobs:
...
+              return node
+
+          with open(sys.argv[1]) as f:
+              head = strip_descriptions(json.load(f))
+          with open(sys.argv[2]) as f:
+              base = strip_descriptions(json.load(f))
+          sys.exit(0 if head == base else 1)
```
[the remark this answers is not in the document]
adtyavrdhn: Fixed in 76adb13. Schemas are compared through canonical JSON serialization, so boolean and integer values remain distinct. A regression test covers true versus 1.

dsfaccini:
@pydanty just a question, can this PR close https://github.com/pydantic/genai-prices/issues/260#issuecomment-5430993304

README.md
```diff
@@ -35,13 +35,13 @@ The following providers are currently supported:
...
 - [AWS Bedrock](prices/providers/aws.yml) - 86 models
-- [Microsoft Azure](prices/providers/azure.yml) - 22 models
+- [Microsoft Azure](prices/providers/azure.yml) - 23 models
 - [Cerebras](prices/providers/cerebras.yml) - 7 models
 - [Cohere](prices/providers/cohere.yml) - 9 models
 - [Deepseek](prices/providers/deepseek.yml) - 7 models
 - [Doubleword](prices/providers/doubleword.yml) - 16 models
 - [Fireworks](prices/providers/fireworks.yml) - 32 models
-- [Google](prices/providers/google.yml) - 51 models
+- [Google](prices/providers/google.yml) - 53 models
```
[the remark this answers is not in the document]
adtyavrdhn: The generated count is correct. `get_providers_yaml()["google"].provider.models` contains 53 records, and `google.yml` has 53 top-level model IDs. `make build` also reports README.md as up to date, so no change is needed.

PR #468, opened 2026-07-16
Refresh Avian provider: add DeepSeek, GLM, MiniMax and Kimi pricing

ayaangazali:
Avian's lineup changed completely since avian.yml was last touched. Their pricing page (https://avian.io/pricing/) now lists DeepSeek V4 Flash/Pro, DeepSeek V3.2, GLM-5, GLM-5.1, MiniMax M2.5 and Kimi K2.5/K2.6, each with a cache price. The four Meta Llama entries no longer appear there, so I marked them deprecated rather than deleting them.

Closes #322

Two judgment calls worth a sanity check:
- the pricing table's column is labeled just "Cache", I mapped it to cache_read_mtok
- model ids follow Avian's documented request format (their site shows model="DeepSeek-V3.2" in the code example); only V3.2 appears verbatim, the others follow the same hyphenated pattern

Gates: make build / lint / typecheck / test all green (coverage stays 100%), npm run ci green, pre-commit all-files green.

---

quick note: I'm a college freshman getting into OSS. I put this together with Claude Code helping me along and checked every number against the pricing page myself. happy to fix anything that looks off :)

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/468?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #469, opened 2026-07-18
Refresh Cerebras provider: add Gemma 4 31B and GLM 4.7, deprecate delisted models

ayaangazali:
Cerebras rotated their inference lineup: the docs now list gpt-oss-120b, gemma-4-31b and zai-glm-4.7, while llama-3.3-70b, llama3.1-8b and qwen-3-32b are gone from the model docs entirely. This adds the two new models (gemma-4-31b at $0.99/$1.49, zai-glm-4.7 at $2.25/$2.75 per MTok, both verified against their official model doc pages, which I also added to pricing_urls) and marks the three delisted ones deprecated with their last known prices kept, same pattern as #468. gpt-oss-120b is unchanged at $0.35/$0.75 so it just gets a prices_checked bump. One heads up: Cerebras has scheduled GLM 4.7 for deprecation on 2026-08-17, noted that in its price_comments.

Gates run locally: make build, lint, typecheck, test (100% coverage), full pre-commit, npm run ci, all green.

*quick note: I'm a college freshman getting into OSS. I built this with Claude Code and double checked every number against the Cerebras docs myself. happy to fix anything I got wrong :)*

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/469?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kludex:
Thanks - Gemma 4 31B ($0.99/$1.49) and GPT OSS 120B ($0.35/$0.75) still match the Cerebras model docs today, and the three delisted models are still absent from the catalog. I merged main into your branch and made two updates: GLM 4.7 was deprecated on 2026-08-17 as scheduled (its doc page is now 404), so it is marked `deprecated: true` with the changelog as the reference instead of the dead link; and `tests/dataset/usages.json` was regenerated since the `zai-glm-4.7` cassette rows now resolve to a priced model.

PR #474, opened 2026-07-21
fix: match compact dated model refs from LiteLLM/OpenRouter

### LINKED ISSUE #333: [LiteLLM/OpenRouter] pricing lookup fails for compact dated model refs

JoshKappler:
Fixes #333. A compact dated model ref like `gpt-5.2-20251211` (which LiteLLM and OpenRouter emit) fails the lookup because the price data only aliases the dashed `gpt-5.2-2025-12-11`. I made model matching retry with the compact date rewritten to the dashed form, but only as a fallback after the ref fails to match as-is, so models that match on the compact date form (like the Bedrock/Anthropic `contains` clauses) are left alone. The date regex is bounded to a valid year, month, and day, and the same change is mirrored in the Python and JS packages. Tested with `pytest tests/test_model_matching.py` (260 passed) and `npm test` in packages/js (634 passed), plus ruff, basedpyright, and tsc clean.

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/474?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/js/src/__tests__/matcher.test.ts
+++ b/packages/js/src/__tests__/matcher.test.ts
@@ -4,7 +4,7 @@ import { describe, expect, it } from 'vitest'
 import type { Provider } from '../types'
 
 import { data } from '../data'
-import { matchLogic, matchModel, matchModelWithFallback, matchProvider } from '../engine'
+import { matchLogic, matchModel, matchModelWithFallback, matchProvider, normalizeCompactDatedRef } from '../engine'
 
 const actualProviders = data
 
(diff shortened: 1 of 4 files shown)


dsfaccini:

Superseded by the merged consolidation in #616 (@dsfaccini). Josh’s original implementation commit and authorship were preserved there, and the consolidated PR credits #474 (@JoshKappler) explicitly. Thank you for the contribution!

PR #478, opened 2026-07-21
Optimize registry-backed pricing hot paths

alexmojaki:
## Summary

- precompute immutable Python registry key indexes instead of rebuilding key collections
- validate already-resolved priced units in Python and JavaScript
- collect each model price and its unit once per calculation, then reuse it for validation, decomposition, tier detection, and aggregation
- preserve request pricing, tiered pricing, custom Python subclasses, custom JavaScript objects, explicit registries, and existing error semantics

## Stack

This PR is stacked on #351 (`codex/data-driven-unit-registry-spec`) and contains only the residual Phase 1 stateless hot-path cleanup.

## Validation

- `make lint typecheck`
- `uv run pytest` — 760 passed, 41 xfailed
- `npm run test --workspace=packages/js` — 733 passed, 1 skipped
- `npm run build --workspace=packages/js`

### DIFF
--- a/packages/python/genai_prices/types.py
+++ b/packages/python/genai_prices/types.py
@@ -700,25 +700,27 @@ class ModelPrice:
     def calc_price(self, usage: AbstractUsage) -> CalcPrice:
         """Calculate the price of usage in USD with this model price."""
         from genai_prices.units import _get_registry  # pyright: ignore[reportPrivateUsage]
-        from genai_prices.validation import validate_model_price
+        from genai_prices.validation import validate_priced_units
 
         registry = _get_registry()
-        validate_model_price(_collect_effective_model_price_keys(self, registry), registry)
+        resolved_prices = _collect_resolved_model_prices(self, registry)
+        validate_priced_units(tuple(unit for unit, _ in resolved_prices), registry)
 
         usage_data = Usage.from_raw(usage)
-        priced_units = _collect_model_price_units(self, registry)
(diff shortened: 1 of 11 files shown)


packages/python/genai_prices/types.py
```diff
@@ -812,29 +810,47 @@ def calc_unit_price(
...
 
 
-def _model_price_uses_tiered_prices(model_price: ModelPrice, registry: UnitRegistry) -> bool:
-    return any(
-        isinstance(getattr(model_price, price_key), TieredPrices)
-        for price_key in _iter_effective_model_price_keys(model_price, registry)
+def _collect_resolved_model_prices(
+    model_price: ModelPrice, registry: UnitRegistry
+) -> tuple[tuple[UnitDef, Decimal | TieredPrices], ...]:
+    stored_prices = [
```
alexmojaki, line 822: **[P2] Preserve dynamically supplied subclass prices**

This collector only enumerates `model_price.__dict__`, whereas the previous implementation resolved every registered key with `getattr`. Registered prices supplied by a subclass property, class attribute, slot, or custom `__getattr__` are now silently omitted. For example, a subclass whose `input_mtok` property returns `Decimal('2')` calculates zero for one million input tokens. Consider retaining the fast `__dict__` path for exact `ModelPrice` instances while resolving registered attributes once for subclasses.

alexmojaki (commented):
The hot-path cleanup looks sound overall, but it introduces one custom-subclass regression that can silently undercharge. Details inline.

PR #515, opened 2026-07-27
feat: price Anthropic cache writes by TTL

alexmojaki:
## Stack

PR 9 of 9 (final). Base PR: #514 (`codex/unit-registry-08-dimension-requirements`).

## Scope

Limited to Anthropic cache-write TTL extraction and pricing: aggregate cache writes retain the default five-minute rate, nested five-minute and one-hour usage is extracted into TTL descendants, one-hour cache writes use the published premium rate, and generated provider data plus Python/JavaScript parity regressions are updated.

## Out of scope

No additional unit vocabulary, provider pricing changes outside Anthropic cache-write TTL, runtime metadata changes, or v1 artifact changes.

## Verification

- Verified aggregate fallback when TTL detail is absent.
- Verified five-minute/one-hour extraction, decomposition, CLI labels, and Python/JavaScript parity when detail is present.
- Ran `make build` twice; the second build was clean.
- Ran Python lint, strict typechecking, and the full suite: 791 passed, 41 expected failures, 100% coverage.
- Ran JavaScript lint, typechecking, and the full suite: 742 passed, 1 skipped.
- Audited every donor-changed path against `14f7802..bef1eec`; all are represented in the stack.
- Confirmed Claude Opus 5 remains present, generated outputs are clean, and all four v1 artifacts remain byte-identical to `bcb2c40`.


tests/test_cli.py
```diff
@@ -493,7 +493,7 @@ def test_calc_table_split_columns_rich_prices(monkeypatch: pytest.MonkeyPatch, c
     assert cli_logic(['calc', '--table', '--input-tokens', '1000', tiered_ref, requests_ref]) == 0
     out, err = capsys.readouterr()
     assert '(+tiers)' in out
-    assert 'Requests/K' in out
+    assert '/ K requests' in out
```
alexmojaki, line 496: how did this change here, without CLI changes

alexmojaki: (AI) Adding the 1-hour cache-write price adds another model-price column. At `COLUMNS=200`, that pushes the table over `_should_split_model_price_columns`' width threshold, so the existing CLI switches to its combined “Model Prices” layout. In that layout the requests price is rendered as `$12 / K requests` rather than using a separate `Requests/K` column header; no CLI implementation changed.

prices/providers/anthropic.yml
```diff
@@ -77,6 +83,7 @@ models:
     prices:
       input_mtok: 0.8
       cache_write_mtok: 1
+      cache_write_1h_mtok: 1.6
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 9aa3b00

Issue #540, opened 2026-07-31
Should the v2 schema freeze be enforced in CI rather than left as a maintainer responsibility?


dsfaccini:
This issue was posted by Claude Code using claude-opus-5 on behalf of David.

## The constraint

`specs/data-driven-unit-registry/phase-1-static-unit-registry-release/code-spec.md` states it plainly:

> Before Phase 2 exists, price releases may update both v2 variants only when every price key and extractor destination is accepted by their checked-in schemas; adding a unit requires v3 instead.

`AGENTS.md` restates it and is honest about the enforcement status:

> **Adding a unit to `prices/units.yml` is a v3 change, not a v2 price update.** [...] Nothing currently blocks this; the constraint is stated in [the code-spec] and is a maintainer responsibility.

## Why it matters

`prices/src/prices/build.py` writes `prices/new_data/v2/data.schema.json` from the registry's allowed price and extractor keys. So adding a unit to `prices/units.yml` and running `make build` regenerates the published v2 schema with the new key in it — silently widening the contract that already-released clients validate against.

Clients that haven't upgraded warn-and-drop unknown price keys rather than failing, which surfaces as silently under-priced usage in production rather than as an error anyone notices.

## Why the existing checks don't catch it

The pre-commit `build` hook (which CI runs via `pre-commit run --all-files`) verifies that the generated artifacts *match* `units.yml`. It cannot distinguish "regenerated because a price changed" from "regenerated because the vocabulary grew" — both are a clean, consistent build. The schema diff is the only signal, and it lands in a file most reviewers scroll past.

## Question

Should this become a CI check rather than a maintainer responsibility? A `v2-schema-frozen` job could pin the expected `prices/new_data/v2/data.schema.json` (and `data_slim.schema.json`) — by digest or by an explicit checked-in copy — and fail the PR when the regenerated schema differs, with the failure message pointing at the v3 requirement.

The tradeoff is that a deliberate v3 cut then requires updating the pin in the same PR, which is arguably the point — it forces the decision to be explicit and reviewable instead of implicit in a large generated diff.

Worth deciding either way, since the current state is a documented rule with nothing behind it.

Related: #539.

alexmojaki:
I'm just hoping to get phase 2 done soon (it shouldn't be that big) and that should solve this. I've also tried to include a broad unit vocabulary so we shouldn't need to add new units any time soon.

Kludex:
[...]

PR #531, opened 2026-07-30
Update OpenAI gpt-5.6 family prices (terra & luna reduced)

dsfaccini:
This pull request was posted by Pi on behalf of David.

OpenAI reduced prices for the gpt-5.6 family. `gpt-5.6-terra` and `gpt-5.6-luna` dropped significantly; `gpt-5.6-sol` is unchanged but re-verified against the current pricing page.

All three use the same tiered structure: short-context base price, and a long-context tier that kicks in above **272K input tokens** (2× input, 1.5× output). Cache writes are billed at 1.25× the uncached input rate; cached input is input ÷ 10.

### Changes (per 1M tokens, short / long)

**gpt-5.6-terra**
| field | before | after |
| --- | --- | --- |
| input | 2.5 / 5 | **2.0 / 4** |
| cache_write | 3.125 / 6.25 | **2.5 / 5** |
| cache_read | 0.25 / 0.5 | **0.2 / 0.4** |
| output | 15 / 22.5 | **12 / 18** |

**gpt-5.6-luna**
| field | before | after |
| --- | --- | --- |
| input | 1.0 / 2 | **0.2 / 0.4** |
| cache_write | 1.25 / 2.5 | **0.25 / 0.5** |
| cache_read | 0.1 / 0.2 | **0.02 / 0.04** |
| output | 6 / 9 | **1.2 / 1.8** |

**gpt-5.6-sol** — unchanged (input 5/10, cache_write 6.25/12.5, cache_read 0.5/1, output 30/45), re-verified.

`prices_checked` updated to `2026-07-30` for all three.

### Verification
- Source: https://developers.openai.com/api/docs/pricing (cross-checked against the per-model pages, e.g. https://developers.openai.com/api/docs/models/gpt-5.6-terra which states the >272K tier rule and 1.25× cache-write multiplier).
- `make build-prices` → regenerated `prices/new_data/v2/data.json`, `data_slim.json`, and package data (`genai_prices/data.py`, `packages/js/src/data.ts`).
- `make test` → 757 passed, 41 xfailed, **100% coverage**.
- `make lint typecheck format` → clean.

### Deliberate scope decisions
- Updated `gpt-5.6-sol`'s `prices_checked` and `price_comments` reference even though its prices didn't change, so the whole family is consistent and points at the per-model docs (which carry both the tier rule and the cache-write rule) rather than the generic caching guide.
- Did not touch `gpt-5.5` or earlier families — their prices still match the OpenAI pricing page.

[...]

PR #542, opened 2026-07-31
Collapse generated diffs, pin the frozen v1 snapshots, document price-history preservation

dsfaccini:
This pull request was posted by Claude Code using claude-opus-5 on behalf of David. David reviewed it lightly.

## Why

`.gitattributes` marked four generated files `linguist-generated`, so GitHub collapses their diffs. The other nine generated files were unmarked, and they dominate the reviewable diff of every price PR — #535 shows +162/−66 on `data.ts` and +64/−16 on `data.py` around a 94-line `openai.yml` change.

Marking the rest is only safe if CI, rather than a reviewer, is what verifies them. For everything `make build` writes, it already is: CI's `pre-commit` job runs `pre-commit run --all-files`, the local `build` hook regenerates all of it, and pre-commit fails when a hook modifies files. Verified by editing `packages/js/src/data.ts` and running the hook:

```
- hook id: build
- files were modified by this hook
```

## The gap this closes

That guarantee does **not** extend to the v1 snapshots, which were already collapsed. No build step writes them any more, so the `build` hook has nothing to compare against. Tampering with `prices/data.json` and running `make build` leaves the edit in place:

```
$ make build && git status --short
 M prices/data.json
```

Collapsed on GitHub, invisible to the build hook — an edit reaches `main` unchecked today.

`tests/test_frozen_v1_data.py` pins the sha256 of all four v1 artifacts (both payloads and both schemas — the schemas were never marked either, and are equally frozen). Confirmed to fail on a tampered payload:

```
AssertionError: data.json is a frozen v1 compatibility snapshot and must not change.
Publish new data under prices/new_data/ instead. If this edit really is intentional,
update the digest in FROZEN_V1_DIGESTS in the same commit.
```

The digests live in an ordinary test file, so a deliberate change has to appear in a diff nobody's tooling collapses.

## Changes

- `.gitattributes` — marks the nine remaining generated artifacts, grouped by what verifies each group, with that stated in comments
- `tests/test_frozen_v1_data.py` — sha256 pin for the four frozen v1 artifacts
- `AGENTS.md` — records what enforces what, so the next agent doesn't have to re-derive it

## Deliberate scope decisions

- **The v2 schema freeze is not addressed here.** `AGENTS.md` says adding a unit to `units.yml` silently widens the published v2 schema and that nothing blocks it. Still true — the `build` hook verifies the schema *matches* `units.yml`, which a widening does. That is a contract guard, not a tamper check, and is tracked separately in #540.
- **No JS test added.** The change is `.gitattributes`, a Python test, and docs; no JS behaviour is touched.
- **v1 artifacts left byte-identical.** The point is to pin them, not to refresh them.

## Verification

- `make lint`, `make typecheck` — pass
- `make test` — all pass, 100% coverage
- `tests/test_frozen_v1_data.py` — passes clean, fails on a tampered payload (shown above)

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/542?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/skills/add-price-model/SKILL.md
+++ b/.claude/skills/add-price-model/SKILL.md
@@ -1,10 +1,11 @@
 ---
 name: add-price-model
 description: >-
-  Add a new LLM model (or provider) to genai-prices pricing data. Use when asked to add/update
-  pricing for a model — e.g. "add grok 4.5", "add the new Claude", "update openai o5 prices". Covers
-  sourcing prices, probing OpenRouter for undocumented dated snapshot IDs, editing the provider YAML,
-  building, verifying resolution, and opening the PR.
+  Add a new LLM model (or provider) to genai-prices pricing data, or change the price of one that is
+  already there. Use when asked to add/update pricing for a model — e.g. "add grok 4.5", "add the new
+  Claude", "update openai o5 prices", "provider X cut its prices". Covers sourcing prices, probing
+  OpenRouter for undocumented dated snapshot IDs, editing the provider YAML, preserving price history
+  across a rate change, building, verifying resolution, and opening the PR.
 ---
 
 # Add a model to genai-prices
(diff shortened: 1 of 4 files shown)


dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David. David reviewed it lightly.

Pushed d6a7441, which widens this PR beyond the original `.gitattributes` scope — flagging it explicitly rather than leaving it in the diff.

## What was added

Guidance so an agent updating an existing model's price does not repeat #531, where the `prices:` block was overwritten in place and every Luna/Terra request from July 9–29 was retroactively re-priced at the reduced rate. #535 is undoing that now.

- `AGENTS.md` § Pricing Data — the rule: never overwrite a `prices:` block on a provider rate change; add a dated conditional price entry, unconstrained entry first, dated entry last (both engines take the last matching entry). The one exception is correcting a value that was already wrong when written.
- `.claude/skills/add-price-model/SKILL.md` — new § 4b with the YAML shape, the ordering rule and why it matters, the `start_date`-vs-`prices_checked` distinction, a boundary-check command, and the requirement to pin both sides in `tests/test_price_calc.py`.
- The skill's `description` now names rate changes. It already claimed "update openai o5 prices" as a trigger but documented only adding a new model, so the case that actually went wrong had no guidance behind it.

## Verification

The YAML example in the skill is not illustrative — it was run through `_providers_from_raw` and resolves as documented:

```
(2026, 7, 29) -> input_mtok 1
(2026, 7, 30) -> input_mtok 0.2
```

The boundary-check command in § 4b was run against real data before being written down.

## Why it is in this PR rather than its own

Both halves address the same failure: a wrong edit reaching `main` because nothing was positioned to catch it. This PR stops reviewers from having to read generated files, so the guidance that keeps the *source* YAML correct belongs with it. Say the word if you would rather it were split — the commits are separate (91f4bc1 / d6a7441) and it splits cleanly.

Issue #494, opened 2026-07-22
`DataSnapshot.find_provider` does a full linear regex scan on every call and caches nothing


dsfaccini:
This issue was posted by Claude Code using claude-opus-4-8 on behalf of David.

David approved filing this and the technical position it takes. He did not review the wording; the measurements and the reading of the code are mine.

`DataSnapshot.find_provider` in `genai_prices/data_snapshot.py` resolves a provider by linearly walking `self.providers` and running `re.match(provider.api_pattern, provider_api_url)` (or `find_provider_by_id`) until one matches. The result depends only on `(provider_id, provider_api_url)` and the snapshot, but nothing is memoized, so the scan is repeated on every call.

`find_provider_model`, immediately above it in the same class, memoizes its result through `self._lookup_cache`. `find_provider` does not.

## Why it shows up

Pydantic AI calls `Provider.extract_usage` through `RequestUsage.extract`, which tries up to three candidate identities in order:

```python
for provider_id, provider_api_url in [(None, provider_url), (provider, None), (provider_fallback, None)]:
    try:
        provider_obj = get_snapshot().find_provider(None, provider_id, provider_api_url)
        _model_ref, extracted_usage = provider_obj.extract_usage(data, api_flavor=api_flavor)
        ...
```

So every request pays at least one scan, and anything on a custom or gateway base URL (which matches no `api_pattern`) pays a full failed scan plus a fallback scan. On streaming providers whose usage metadata is cumulative, this repeats per chunk: Gemini reports `usage_metadata` on every streamed chunk, so a 100 chunk stream runs the scan 100 times for an answer that cannot change.

## Measurements

genai-prices 0.0.71, 34 providers in the bundled snapshot, CPython 3.13.3, macOS. Timed with `timeit` over 20k iterations per call, four independent runs in an isolated environment; the ranges below are min to max across those runs.

| call | µs |
|---|---:|
| `find_provider(None, 'anthropic', None)` (first in the list) | 0.1 to 0.3 |
| `find_provider(None, None, 'https://generativelanguage.googleapis.com')` | 2.2 to 2.9 |
| `find_provider(None, None, 'https://api.openai.com/v1')` | 5.6 to 7.7 |
| `find_provider(None, '<unknown provider id>', None)`, raises `LookupError` | 5.1 to 7.0 |
| `find_provider(None, None, '<unmatched base_url>')`, raises `LookupError` | 7.2 to 13.5 |
[...]

PR #560, opened 2026-08-10
Fix order-dependent price lookup: cache write key must match its read key

Hrafz:
## Summary

`DataSnapshot.find_provider_model` reads and writes `_lookup_cache` under two different
keys, so a `calc_price` call with no provider returns a different price depending on what
ran before it in the same process.

The provider-supplied branch reads:

```python
if provider:
    if provider_model := self._lookup_cache.get((provider.id, None, model_ref)):
        return provider_model
```

but the single write further down keys off the *parameters*, not `provider.id`:

```python
self._lookup_cache[(provider_id, provider_api_url, model_ref)] = ret = provider, model
```

`extract_usage` resolves the provider itself and calls
`find_provider_model(model_ref, provider, None, None)`, so on that path `provider_id` and
`provider_api_url` are both `None` and the write lands on `(None, None, model_ref)` —
exactly the key the provider-less branch reads. Two consequences, and they compound:

1. A provider-qualified `extract_usage` poisons the entry that a later provider-less
   `calc_price` for the same `model_ref` finds, so the later call skips `find_provider`
   and inherits the earlier call's provider.
2. The read key on the provider-supplied branch is never written by anything, so that
   branch's memo never hits. It is currently decorative.

## Reproduction

`claude-opus-4-6` is defined by both `anthropic` and `google`, at different rates. Same
call, same arguments, two answers:

```python
import genai_prices
from genai_prices import Usage

usage = Usage(input_tokens=1_000_000, output_tokens=1_000_000)

# in a fresh process
genai_prices.calc_price(usage, model_ref='claude-opus-4-6')
#=> provider 'anthropic', total_price 30.00

# in a fresh process, with one unrelated extraction first
genai_prices.extract_usage(
    {'modelVersion': 'claude-opus-4-6',
     'usageMetadata': {'promptTokenCount': 1, 'candidatesTokenCount': 1}},
    provider_id='google',
)
genai_prices.calc_price(usage, model_ref='claude-opus-4-6')
#=> provider 'google', total_price 47.50
```

**$30.00 and $47.50 for identical arguments** — a 58% difference decided by call order.
Both runs are on `main` at `5819739`. With `claude-opus-5` the provider still flips but
the two rates happen to agree, which is presumably why this has gone unnoticed: the
symptom is only a wrong price when the two providers disagree.

A pricing library ought to be a pure function of its arguments, and `calc_price` with an
explicit `provider_id` already is — only the provider-less path is affected.

## Fix

Compute the cache key once per branch and use it for both the read and the write:

```python
if provider:
    cache_key = (provider.id, None, model_ref)
    if provider_model := self._lookup_cache.get(cache_key):
        return provider_model
else:
    cache_key = (provider_id, provider_api_url, model_ref)
    if provider_model := self._lookup_cache.get(cache_key):
        return provider_model

    provider = self.find_provider(model_ref, provider_id, provider_api_url)

if model := provider.find_model(model_ref, all_providers=self.providers):
    self._lookup_cache[cache_key] = ret = provider, model
    return ret
```

+8/−3 in `packages/python/genai_prices/data_snapshot.py`. No cache entry is dropped and
nothing is invalidated more often — the provider-supplied branch simply starts writing
the key it already reads.

## Python/JS parity

`AGENTS.md` asks that matching changes land on both sides. **This one is Python-only, and
that is the point:** `packages/js` has no lookup cache at all — `matchProvider` in
`engine.ts` re-resolves on every call, so the JS implementation is already order-independent
and produces the `anthropic` answer in both scenarios above. The Python behaviour was the
divergence; this removes it. I checked for a JS equivalent before concluding that
(`grep -rn 'cache' packages/js/src`, and `matchProvider` / `findProviderById` read in full).

## Tests

`tests/test_snapshot_cache.py`, two tests, both failing on `main` and passing with the fix:

- a provider-less `calc` is unchanged by a preceding provider-qualified `extract_usage`
  for the same `model_ref` — asserted on both `provider.id` and `total_price`;
- the provider-supplied branch's read key is populated by its own write. Pre-fix this
  fails with the cache contents printed, showing the entry sitting on
  `(None, None, 'claude-opus-4-6')`.

Neither test hardcodes a rate — they compare the two call orders against each other — so a
price update does not touch them. Each builds its own `DataSnapshot`, so no state leaks
between them or into the rest of the suite.

Verified on `main` at `5819739`: `pytest -q` goes from 785 to 787 passed (41 xfailed
unchanged), `ruff check` and `ruff format --check` clean, `make typecheck` 0 errors, and
`coverage` reports 100% on both changed files.

## Related, not duplicates

#494 reports that `find_provider` does a linear regex scan and caches nothing. Different
method, opposite complaint — but anyone adding a cache there will be working next to this
key-symmetry problem, so the two are worth reading together.

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/560?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


adtyavrdhn:
Thanks @Hrafz

adtyavrdhn (approved):

PR #572, opened 2026-08-19
Reject mapping usage inputs instead of pricing them at zero

### LINKED ISSUE #516: Reject mapping usage inputs instead of silently returning a zero price
## Problem

In #509, `Usage.from_raw(...)` normalizes usage by reading registered attributes with `getattr`. A plain mapping therefore contributes no recognized usage values, even when it contains valid usage keys.

Passing a mapping to standard pricing now succeeds with an empty usage object and returns a zero price:

```python
from decimal import Decimal

from genai_prices.types import ModelPrice

result = ModelPrice(input_mtok=Decimal("1")).calc_price(
    {"input_tokens": 1_000_000}
)

assert result["total_price"] == Decimal("0")
```

Mappings were not previously a supported usage input, but they failed rather than silently producing a plausible undercharge. The new behavior is explicitly asserted by `test_usage_from_raw_ignores_mapping_keys`.

## Expected behavior

Either:

- support recognized keys in mapping inputs; or
- preferably, while mapping compatibility remains out of scope, reject mappings with a clear `TypeError` or `ValueError`.

Unsupported input must not silently calculate a zero price.

## Additional consideration

The same normalization policy silently ignores misspelled or unrelated attributes on generic raw objects. It may be worth requiring at least one recognized reported usage value when a non-request price needs usage, while preserving valid request-only and explicitly empty `Usage()` calculations.

Related: #509

arthi-arumugam-git:
Fixes #516.

## What happens now

On current main the same usage prices differently depending on its shape:

```python
price = ModelPrice(input_mtok=Decimal('1'))

price.calc_price({'input_tokens': 1_000_000})                      # 0
price.calc_price(SimpleNamespace(input_tokens=1_000_000))          # 1
```

A million input tokens costs nothing if you pass a dict, because `Usage.from_raw` reads values with `getattr` and a mapping has no such attributes, so it builds an empty `Usage` and everything downstream prices honestly against nothing.

## The change

`Usage.from_raw` now raises `TypeError` for mappings, naming the type it received. That is the second option in the issue, rejecting rather than supporting mapping keys, since mapping compatibility is out of scope.

The blast radius is small: `from_raw` has one internal caller, `ModelPrice.calc_price`, so the only behaviour that changes is an input that previously produced a silent undercharge.

## About the existing test

`test_usage_from_raw_ignores_mapping_keys` asserted the current behaviour, so I replaced it rather than worked around it. In its place:

- `test_usage_from_raw_rejects_mappings`, the same input now expecting the `TypeError`
- `test_calc_price_rejects_mapping_instead_of_pricing_it_at_zero`, which pairs the rejection with the assertion that the same usage as an object still prices at 1, so only the unsupported shape changed

Both fail on main.

## One thing I noticed, for your call rather than mine

The JS package types usage as `Record<string, number | undefined>`, so a plain object is its native and only usage input and the same `{input_tokens: 1_000_000}` prices correctly there. Since `AGENTS.md` says the two implementations must stay behaviourally identical, this fix makes Python raise where JS accepts. That seemed like the right reading of the issue, which scopes mapping support out for Python, but if you would rather close the gap the other way I am happy to redo it as mapping support instead.

## Tests

`tests/test_usage.py` 48 passed, and the suites that collect in a plain venv here are green: `test_extract_usage` 74, `test_decompose` 20, `test_custom_prices` 11, `test_lifecycle` 11, `test_end_to_end` 7. I could not run the data-pipeline suites locally, they need the `prices` build tooling.

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/572?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/python/genai_prices/types.py
+++ b/packages/python/genai_prices/types.py
@@ -222,9 +222,12 @@ class Usage:
         # would price at zero however many tokens it names. Mappings are not a
         # supported usage input, so say so rather than undercharge silently.
         if isinstance(obj, Mapping):
+            # isinstance narrows to Mapping[Unknown, Unknown], so name the type through a
+            # cast to keep the type checker from seeing partially unknown arguments.
+            unsupported = cast('Mapping[Any, Any]', obj)
             raise TypeError(
                 'Mappings are not supported as usage input. '
-                f'Pass a Usage instance or an object with usage attributes, not {type(obj).__name__}.'
+                f'Pass a Usage instance or an object with usage attributes, not {type(unsupported).__name__}.'
             )
 
         values: dict[str, UsageValue] = {}
--- a/packages/python/genai_prices/types.py
+++ b/packages/python/genai_prices/types.py
@@ -218,6 +218,15 @@ class Usage:
             obj._reported_values()
             return obj
 
+        # Usage values are read with getattr, so a mapping contributes nothing and
+        # would price at zero however many tokens it names. Mappings are not a
+        # supported usage input, so say so rather than undercharge silently.
+        if isinstance(obj, Mapping):
+            raise TypeError(
+                'Mappings are not supported as usage input. '
+                f'Pass a Usage instance or an object with usage attributes, not {type(obj).__name__}.'
+            )
+
         values: dict[str, UsageValue] = {}
         for key in _reported_usage_keys():
             value = _raw_usage_value(obj, key)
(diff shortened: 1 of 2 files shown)


Kludex:
Closing this approach. I’ll implement the issue separately with a narrower input-shape check.

PR #575, opened 2026-08-19
Stabilize the uv lockfile writer

ddanielcruzz:
## Summary

- require `uv >=0.12.1` and install exactly `uv 0.12.1` in CI and agent setup
- align both `uv_build` constraints with the 0.12 release and accept the one-time canonical lockfile marker normalization
- use `uv version` for release bumps, followed by a locked full-workspace sync that verifies installation without rewriting the lockfile

## Why

A routine package-version bump could produce a large unrelated-looking `uv.lock` diff when it was run with a different `uv` lockfile writer. The repository also allowed automation to install an unspecified latest version, so identical commits could be checked with different `uv` behavior over time.

This makes the supported minimum explicit, gives automation one canonical writer, and keeps the release script's installation check while preventing that check from changing the freshly generated lockfile.

A simulated `0.1.3` to `0.1.4` release now changes only the four expected version lines in the Python package, JavaScript package, `package-lock.json`, and `uv.lock`.

## Validation

- `uv lock --check`
- `uv sync --locked --all-packages`
- `make all` (837 passed, 41 expected failures, 100% coverage)
- `npm run ci` (1,852 passed, 1 skipped)
- disposable-clone `./uprev.sh 0.1.4` release simulation

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/575?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/copilot/actions-setup-steps.yml
+++ b/.github/copilot/actions-setup-steps.yml
@@ -5,6 +5,7 @@ steps:
   - name: Setup UV (Python package manager)
     uses: astral-sh/setup-uv@v6
     with:
+      version: "0.12.1"
       enable-cache: true
 
   - name: Setup Node.js
--- a/.github/workflows/after-ci.yml
+++ b/.github/workflows/after-ci.yml
@@ -15,6 +15,7 @@ jobs:
     steps:
       - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
         with:
+          version: "0.12.1"
           enable-cache: true
           python-version: "3.12"
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
           persist-credentials: false
       - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
         with:
+          version: "0.12.1"
           enable-cache: true # zizmor: ignore[cache-poisoning] -- Job does not produce release artifacts and does not have sensitive permissions
       - run: uv sync  --all-packages
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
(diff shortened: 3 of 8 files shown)


dsfaccini (approved):
> This review was posted by Claude Code using claude-opus-5 on behalf of David.

David reviewed this lightly.

Approving. CI is green across py3.10-3.14, pre-commit, and coverage; the diff is tight and the rationale holds up.

Notes from verification beyond reading the diff:

**The `uv.lock` diff is purely marker normalization.** Filtering out every marker-only line leaves nothing behind — no package, version, or hash changes.

**The `uv_build` bump on the published package is not exercised by this PR's CI.** `release-pypi` is the only job that runs `uv build`, and it runs only on tags. I built `packages/python` under both the old (`>=0.8.10,<0.12`) and new (`>=0.12.1,<0.13`) constraints and compared:

- wheel file list — identical
- `METADATA` — byte-identical
- sdist — one difference: uv_build 0.12 rewrites `pyproject.toml` inside the sdist (inline `authors` becomes a `[[project.authors]]` table) and ships the pre-rewrite copy as a new `pyproject.toml.orig` entry

That sdist change is cosmetic and semantically identical, so it is not a blocker — recording it here so the next person who diffs a published sdist knows where the extra file came from.

**`uv version --no-sync` does re-lock** (`--no-sync` avoids syncing the virtual environment *after* re-locking), so the `uv sync --locked --all-packages` that follows it in `uprev.sh` is a genuine verification rather than a step that would fail on a stale lockfile.

All seven `setup-uv` call sites received the version pin: five in `ci.yml`, plus `after-ci.yml` and `.github/copilot/actions-setup-steps.yml`.

Two adjacent observations, neither blocking:

- `required-version = ">=0.12.1"` hard-fails every uv invocation for contributors on an older uv. The error names `uv self update`, so it is self-explanatory, but nothing in the repo states the minimum. A line in the dev-setup docs would save a round trip.
- `.github/copilot/actions-setup-steps.yml` still floats on `astral-sh/setup-uv@v6` while the workflows are SHA-pinned to v8.1.0. Pre-existing and out of scope here.

PR #577, opened 2026-08-21
Stop new command-a variants matching the base command-a price

### LINKED ISSUE #471: Cohere: new command-a variants silently priced as command-a via starts_with match
cohere.yml matches Command A with `starts_with: command-a`, which was fine when command-a was the only model in that family. But Cohere's model list (https://docs.cohere.com/docs/models) now has command-a-plus-05-2026, command-a-reasoning-08-2025, command-a-vision-07-2025 and command-a-translate-08-2025 as separate models, and they all silently fall into the command-a entry:

```python
from genai_prices import calc_price, Usage
p = calc_price(Usage(input_tokens=1_000_000, output_tokens=1_000_000), model_ref='command-a-plus-05-2026', provider_id='cohere')
print(p.model.id, p.total_price)  # command-a 12.5
```

so anyone tracking spend on the newer models gets command-a's $2.50/$10 rates with no warning.

I wanted to just PR the missing models with their real prices, but I couldn't find current per-token pricing published anywhere: https://cohere.com/pricing only exposes the legacy table (Command, Command-light, R 03-2024, R+ 04/08-2024) plus Model Vault hourly rates in its server-rendered payload, and the docs model pages don't carry pricing either. Options I can see:

1. add the new models with correct prices (needs a source for the numbers, maybe someone with Cohere access can confirm them)
2. tighten the command-a match (e.g. equals on command-a / command-a-03-2025) so the newer variants raise LookupError instead of silently using wrong prices

happy to do the yml work for either once there's a direction. sorry I couldn't come with the fix directly, I tried to put the refresh PR together first but got stuck on the missing price source (used Claude Code to help dig through the pricing page payload). still learning my way around, so if I've missed an obvious pricing source please point me at it :)

AmirF194:
Root cause: `command-a`'s match clause in `prices/providers/cohere.yml` is `starts_with: command-a`. Cohere has since shipped `command-a-plus-05-2026`, `command-a-reasoning-08-2025`, `command-a-vision-07-2025` and `command-a-translate-08-2025`, none of which have published per-token pricing yet. All four start with `command-a`, so they silently match the base model's rate ($2.50/$10 per Mtok) instead of raising `LookupError` the way an unpriced model is supposed to.

Fix: replace the `starts_with` clause with an explicit `or` of the two model refs it should actually match, `command-a` and `command-a-03-2025` (the one already pinned by `tests/test_model_matching.py`). The new variants now fall through to `LookupError` like any other unpriced model, instead of silently picking up the wrong price.

Regenerated the packaged data with `make build` (v2 JSON, `data.py`, `data.ts`) so the Python and JS packages stay in sync from the same YAML source.

Verification:
- `test_command_a_variants_do_not_match_base_command_a` (new, 4 cases): fails on main (all 4 variants resolve to `command-a`), passes on the branch (all 4 raise `LookupError`), run both ways in a clean `python:3.13-slim` container.
- `make test` (Python, 100% branch coverage) and `npm run ci` (JS) both pass in full, including the existing pinned `command-a-03-2025` case.
- `pre-commit run --all-files` (ruff, basedpyright, prettier, the `build`/`check-for-price-discrepancies`/`inject-providers` hooks) passes clean, no rewrites needed.
- Not checked: whether Cohere publishes real per-token pricing for these four variants in the future; when they do, they should get their own priced entries rather than staying unmatched.

Fixes #471

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/577?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


AmirF194:
Rebased onto current main to clear the conflict and force-pushed (494e951). No other changes, full test suite still green (1177 passed).

AmirF194:
Closing this one: #604 landed the same fix (restricting the command-a match to known model IDs) about 40 minutes before I could rebase past the conflict it introduced. I checked the diffs and they're identical on the source files this PR touched. The only thing left in this branch is a handful of extra regression-test cases for variant IDs #604's own test doesn't cover; not worth a PR on its own. Thanks for getting to it.

PR #583, opened 2026-08-22
Add long-context price tiers for 4 Grok models and both GPT-5.5 models

xyzs996:
## What this changes

Six models were priced with a single flat rate, but their vendors publish a **second, higher rate for long prompts**. Four Grok models and the two GPT-5.5 models now carry `tiers`.

xAI's docs say it plainly: requests whose prompt reaches the listed token threshold are billed at the higher rate **for all tokens in the request**. OpenAI's 272K tier works the same way. That is `TieredPrices`' documented cliff semantics, so no engine change is needed — only data.

| model | threshold | `input_mtok` | `cache_read_mtok` | `output_mtok` |
| --- | ---: | --- | --- | --- |
| `grok-4.5` | 200k | 2 → 4 | **0.30** → 0.60 | 6 → 12 |
| `grok-4.3` | 200k | 1.25 → 2.5 | 0.20 → 0.40 | 2.5 → 5 |
| `grok-4.20` | 200k | 1.25 → 2.5 | 0.20 → 0.40 | 2.5 → 5 |
| `grok-build-0.1` | 200k | 1 → 2 | 0.20 → 0.40 | 2 → 4 |
| `gpt-5.5` | 272k | 5 → 10 | 0.50 → 1 | 30 → 45 |
| `gpt-5.5-pro` | 272k | 30 → 60 | – | 180 → 270 |

Why it matters: reading a cliff as a marginal tier under-states a long-context bill by close to half. A 1M-token prompt on `gpt-5.5` costs $10, not the $7.28 you get by charging the first 272k at the base rate.

## Sources

Every number is double-sourced, and the two sources agree on all six rows:

- Vendor docs: https://docs.x.ai/docs/models (read 2026-08-22) and https://developers.openai.com/api/docs/pricing (read 2026-08-22).
- OpenRouter's live `/api/v1/models`, where the same thresholds ship as `pricing.overrides[].min_prompt_tokens`.

There is also strong corroboration inside this repo: `gpt-5.6-sol` and `gpt-5.4-pro` already carry 272k tiers whose base numbers are identical to the flat `gpt-5.5` / `gpt-5.5-pro` rows this PR converts. The tier was already understood here — these six rows just never got it.

`prices_checked` moves to 2026-08-22 on every touched model, and each gets a `price_comments` line stating the cliff and linking the vendor page.

## One correction beyond adding tiers

`grok-4.5`'s `cache_read_mtok` base was **0.50**; xAI publishes **0.30**. 0.50 is Grok **4.6**'s cache-read price, so this looks like a value copied from the neighbouring model. Corrected here with a note in `price_comments`; the tier above it (0.60) is 2x the corrected base, consistent with every other row in the table.

## Two things flagged, not changed

**`grok-4.20-multi-agent`** is priced `2 / 0.2 / 6` in this repo. Both xAI's docs and OpenRouter say the multi-agent variant costs the same as plain `grok-4.20`: `1.25 / 0.2 / 2.5`. I did not touch it — a flat-number disagreement is a weaker claim than adding a vendor-documented tier, and it deserves its own look rather than riding along in this PR. (Same pattern as the reseller-price observation in #581.)

**`grok-4.6` is missing entirely.** It is xAI's current flagship (`2 / 0.50 / 6`, doubling past 200k), but there is no row for it here: `calc_price(model_ref='grok-4.6')` raises `LookupError: Unable to find model with model_ref='grok-4.6' in x-ai`. Adding a model is more than a price fix, so I left it out of this PR — say the word and I'll open one.

## Boundary: `>` vs "reaches"

`calc_unit_price` resolves tiers with `total_input_tokens > tier.start`, so a prompt of **exactly** 200,000 tokens bills at the base rate here, while xAI's wording ("reaches the listed threshold") puts it at the higher rate. That is a repo-wide convention shared with every existing tiered row, so I kept it rather than making these six rows behave differently — the tests below assert the repo's behaviour, not the vendor's, so the difference is visible rather than silent. Happy to flip it repo-wide in a separate PR if you want the other reading.

## Tests

`tests/dataset/usages.json` pins short-prompt usages, so it cannot catch a missing long-context tier. Added a parametrized test on each side, over the shipped data, for all six models:

- `tests/test_price_calc.py::test_price_long_context_cliff` and `::test_price_long_context_cliff_is_not_marginal`
- `packages/js/src/__tests__/comprehensive.test.ts` → `calcPrice - long-context cliff`

Each model is checked at three points: exactly at the threshold (base rate), one token over (higher rate applied to the **whole** request), and an assertion that one extra token roughly doubles the bill — which fails loudly if anyone re-implements the tier as marginal. A separate case pins `gpt-5.5` at 1M input tokens to $10 and asserts it is *not* the marginal figure.

## Verification

Every generated file was written by `make build` — none hand-edited.

```
make build       # data.py, data.ts, v2/data.json, v2/data_slim.json
make lint        # 61 files already formatted; all checks passed
make typecheck   # 0 errors, 0 warnings, 0 notes
make test        # 860 passed, 41 xfailed, 100.00% coverage; usages.json up to date
npm run ci       # 1859 passed, 1 skipped, 14 files
```

Diff of the generated v2 payload, confirming nothing outside these six models moved:

```python
import json, subprocess
old = json.loads(subprocess.run(['git','show','main:prices/new_data/v2/data.json'],
                                capture_output=True, text=True, check=True).stdout)
new = json.load(open('prices/new_data/v2/data.json'))
flat = lambda d: {(p['id'], m['id']): json.dumps(m, sort_keys=True)
                  for p in d for m in p.get('models', [])}
a, b = flat(old), flat(new)
assert not set(a) ^ set(b)                                  # no models added or removed
print(sorted(k for k in a if a[k] != b[k]))
# [('openai', 'gpt-5.5'), ('openai', 'gpt-5.5-pro'), ('x-ai', 'grok-4.20'),
#  ('x-ai', 'grok-4.3'), ('x-ai', 'grok-4.5'), ('x-ai', 'grok-build-0.1')]
```

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/583?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


xyzs996:
Third-source check, added after opening this: [models.dev](https://models.dev/api.json)'s catalog agrees with every number here, independently of xAI's docs and OpenRouter's API.

| model | this PR | models.dev `cost` + `tiers` |
| --- | --- | --- |
| `grok-4.5` | 2 → 4, **0.30** → 0.60, 6 → 12 | `input 2, cache_read 0.3, output 6` + tier @200k `4 / 0.6 / 12` |
| `grok-4.3` | 1.25 → 2.5, 0.2 → 0.4, 2.5 → 5 | same |
| `grok-4.20` | 1.25 → 2.5, 0.2 → 0.4, 2.5 → 5 | same (both `-reasoning` and `-non-reasoning`) |
| `grok-build-0.1` | 1 → 2, 0.2 → 0.4, 2 → 4 | same |
| `gpt-5.5` | 5 → 10, 0.5 → 1, 30 → 45 | same, tier @272k |
| `gpt-5.5-pro` | 30 → 60, 180 → 270 | same, tier @272k |

Two things this settles:

- **The `grok-4.5` cache-read correction.** models.dev has `cache_read: 0.3` for 4.5 and `0.5` for 4.6 — same split as xAI's docs, so the `0.5` currently in this repo really is 4.6's number on 4.5's row.
- **The `grok-4.20-multi-agent` flag.** models.dev has `grok-4.20-multi-agent-0309` at `1.25 / 0.2 / 2.5`, identical to plain `grok-4.20`. That's now three independent sources against this repo's `2 / 0.2 / 6`. Still not touching it here — happy to send it as its own one-line PR if you'd rather see it separately.

(Unrelated to the numbers, but found in the same pass and reported upstream: models.dev's SDK documents `context_over_200k` as a flat 200K threshold, while 193 of the 357 entries carrying it are actually 256K/272K/512K — anomalyco/models.dev#5277. Worth knowing if anything here ever imports from that field rather than from `tiers`.)

xyzs996:
Sharpening two things in the description while this is unreviewed.

**Sourcing, precisely.** The *"billed at the higher rate for all tokens in the request"* sentence is quoted from xAI. OpenAI's pricing page does **not** state the billing mechanic — it labels the row `gpt-5.5 (<272K context length)` and gives two columns. So "one rate card selected by prompt length" is a direct quote for xAI and an inference from the card layout for OpenAI, corroborated by OpenRouter modelling both the same way (`min_prompt_tokens`, one flat override, no banding). The numbers are quoted; only the mechanic is inferred, and I'd rather say which is which.

**The boundary note has a concrete instance, in this repo, already correct.** Three vendors, two conventions, one token apart:

| vendor | wording | where 200 000 / 272 000 exactly lands |
| --- | --- | --- |
| Google | `prompts <= 200k` / `> 200k` | cheap |
| xAI | prompts that *reach* 200k | dear |
| OpenAI | `(<272K context length)` | dear |

`TieredPrices` resolving with `tokens > tier.start` matches Google exactly — and `gemini-2.5-pro` and `gemini-3.1-pro-preview` are already tiered here, so the convention isn't arbitrary, it's Gemini's, applied repo-wide. It's off by one token for the six rows in this PR and for `grok-4.6` in #584.

I've still matched it rather than special-casing, because a per-model boundary flag is a bigger change than these rows justify and the cost is one token. But it's now a documented heterogeneity rather than an unexamined default: if you'd rather have `>=` per model, say so and I'll do it in a follow-up rather than widening this one.

xyzs996:
Added `grok-4.6` in aada952. It is xAI's currently recommended flagship on their models page and it is missing from `x_ai.yml` altogether, so it matches nothing today — and it has the same two-row shape as the models already in this PR, so it belongs with them rather than in a separate change.

From https://docs.x.ai/docs/models, checked 2026-08-22: 500k context, `$2.00 / $0.50 / $6.00` below 200k prompt tokens and `$4.00 / $1.00 / $12.00` at or above, with the same footnote — a request whose prompt reaches the threshold is billed at the higher rate for every token in it.

Its `$0.50` base cache-read price is also the reason grok-4.5's `price_comments` in this PR calls out that 0.50 was never 4.5's number; the two look like they were transcribed across at some point.

`uv run pytest` 860 passed / 41 xfailed, `npm test` 1859 passed / 1 skipped, ruff format + check clean, prettier clean, and `uv run -m prices collapse` reports "No models combined".

I also opened #586 by mistake covering part of the same file; it is closed now, with everything folded in here.

xyzs996:
Reverted the grok-4.6 commit in 7482098 — grok-4.6 is already #584, which was cut from `main` specifically so it would not collide with this branch, and which has four agreeing sources and a test. I added it here without checking my own open PRs first. Sorry for the churn; this branch is back to what it was at 8545ef4.

So the intended split stands: **this PR** = long-context tiers for the 4 existing Grok models and both GPT-5.5 models, **#584** = the missing grok-4.6 row.

xyzs996:
Pushed the same xAI boundary fix here. Their tables read `< 200k prompt tokens` / `>= 200k prompt tokens` and the footnote says a prompt that *reaches* the threshold is billed high for every token, while tiers fire on `tokens > start` — so `start: 200000` under-billed an exactly-200,000-token prompt. All 12 xAI entries are now `199999`.

OpenAI is left as-is: their page only labels the cheap column "<272K context length", so 272000 itself is undocumented. The cliff cases now carry the first token count billed long rather than the threshold, since the two vendors put the line on different sides.

dsfaccini:

Consolidated in #615, which has now merged. Thank you @xyzs996 — your original commit authorship was preserved in #615 so your contribution remains credited.

PR #586, opened 2026-08-22
xAI: add the documented >=200k long-context tiers, add grok-4.6

xyzs996:
All four current Grok models are priced as two rows on xAI's own models page, and this repo carries only the first row — so every request over 200k prompt tokens is priced at half what xAI charges. `grok-4.6`, which that page recommends for both coding and chat, is missing from the file entirely.

### What the vendor page says

| row label | context | input | cached input | output |
| --- | --- | --- | --- | --- |
| `grok-4.6 (< 200k prompt tokens)` | 500k | $2.00 | $0.50 | $6.00 |
| `grok-4.6 (≥ 200k prompt tokens)` | 500k | $4.00 | $1.00 | $12.00 |
| `grok-4.5 (< 200k prompt tokens)` | 500k | $2.00 | $0.30 | $6.00 |
| `grok-4.5 (≥ 200k prompt tokens)` | 500k | $4.00 | $0.60 | $12.00 |
| `grok-4.3 (< 200k prompt tokens)` | 1M | $1.25 | $0.20 | $2.50 |
| `grok-4.3 (≥ 200k prompt tokens)` | 1M | $2.50 | $0.40 | $5.00 |
| `grok-4.20-0309-reasoning (< 200k prompt tokens)` | 1M | $1.25 | $0.20 | $2.50 |
| `grok-4.20-0309-reasoning (≥ 200k prompt tokens)` | 1M | $2.50 | $0.40 | $5.00 |
| `grok-4.20-multi-agent-0309 (< 200k prompt tokens)` | 1M | $1.25 | $0.20 | $2.50 |
| `grok-4.20-multi-agent-0309 (≥ 200k prompt tokens)` | 1M | $2.50 | $0.40 | $5.00 |

with the footnote:

> Models listed with two rows use long context pricing […] requests whose prompt **reaches** the listed token threshold are billed at the higher rate for all tokens in the request.

— [xAI models & pricing](https://docs.x.ai/docs/models), read 2026-08-22.

That is exactly the cliff semantics `TieredPrices` already documents ("crossing a tier applies that rate to ALL tokens"), so this is a pure data change.

### What this changes

Tiers at `start: 200000` on `grok-4.20`, `grok-4.20-multi-agent`, `grok-4.3` and `grok-4.5`; a new `grok-4.6` entry; `context_window: 1000000` on the two `grok-4.20` entries, which had none.

**Two base prices move as a consequence, and I want to flag them rather than bury them** — both are read off the same rows quoted above, and leaving them would make the entries internally incoherent (a "2×" tier sitting on top of a base it isn't 2× of):

- `grok-4.5` cached input is **$0.30**, not the $0.50 currently in the file. $0.50 is `grok-4.6`'s rate at the same tier. For agent workloads, where cache reads dominate the token mix, this is the number that actually shows up on the bill.
- `grok-4.20-multi-agent` is priced identically to the other `grok-4.20` variants — **$1.25 / $0.20 / $2.50**, not $2 / $0.20 / $6.

If you'd rather take the tier additions alone and handle those two separately, say so and I'll split them out.

**One knock-on effect, flagged because it deletes an id.** Once `grok-4.20-multi-agent` carries the same prices as `grok-4.20`, the repo's own `collapse models` hook merges the two entries, and CI enforces that — my first push failed on exactly this. I let the hook do it rather than adding `collapse: false`. `grok-4.20-multi-agent` stays reachable as a `match` alias on the merged entry, so lookups are unaffected; only the `id` in the data changes. Say the word if you'd rather keep it as its own entry.

### Effect

```
                       before      after
grok-4.6  400k in       (absent)   $1.6120
grok-4.5  400k in       $0.8060    $1.6120
grok-4.3  400k in       $0.5025    $1.0050
grok-4.20 400k in       $0.5025    $1.0050
```

### One thing this does *not* fix

`calc_unit_price` selects a tier with `if total_input_tokens > tier.start`, so at **exactly** 200,000 prompt tokens these models still bill at the base rate — while xAI's wording is "≥ 200k" / "reaches the listed token threshold", i.e. that token is over the line.

I deliberately left that alone, because the strict `>` is right for at least one other provider and fixing it needs a schema decision that's yours to make. Google documents the opposite convention for the same threshold number:

> $2.00, prompts <= 200k tokens $4.00, prompts > 200k tokens

— [Gemini API pricing](https://ai.google.dev/gemini-api/docs/pricing), read 2026-08-22.

So a 200,000-token prompt is under the line on Google and over it on xAI — same number, opposite answer — and one global comparison operator cannot be right for both. Happy to open a separate issue for that if it's useful; it would need something like a per-tier inclusivity flag.

### Checks

- `uv run pytest` — 853 passed, 41 xfailed
- `npm test` — 1852 passed, 1 skipped (all 1577 dataset assertions unchanged)
- `uvx pre-commit run --all-files` equivalents: `make build` and `uv run -m prices collapse` are both clean and idempotent on the final tree


xyzs996:
Closing this — I duplicated my own earlier PR and only spotted it after CI went green here. Sorry for the noise.

#583 already adds the same `start: 200000` tiers to `grok-4.20`, `grok-4.3` and `grok-4.5` from the same source page, including the same `cache_read_mtok` 0.5 → 0.3 correction on grok-4.5, and it also covers `grok-build-0.1`, both `gpt-5.5` models in `openai.yml`, and the test additions. Two open PRs editing `x_ai.yml` would just conflict with each other.

The one thing this PR had that #583 did not is the missing `grok-4.6` entry, so I have moved it there as a follow-up commit rather than keeping a second PR alive. Everything worth reviewing is now in #583.

One piece I am deliberately **not** carrying over: this PR also corrected `grok-4.20-multi-agent`'s base prices to `$1.25 / $0.20 / $2.50`, which made it price-identical to `grok-4.20` and caused the collapse hook to merge the two ids. That is a base-price change plus an id removal, which does not belong in a PR about long-context tiers — I will raise it separately so it can be reviewed on its own terms.

PR #596, opened 2026-08-24
Add cache_write_tokens mapping to Azure's responses/chat extractors

kecbigmt:
## Summary

The `openai` provider's `chat` and `responses` `UsageExtractor`s already map `*_tokens_details.cache_write_tokens` to `cache_write_tokens`, but the `azure` provider (`api_pattern` matches `*.openai.azure.com`, `*.azure-api.net`, `*.cognitiveservices.azure.com`) is missing the identical mapping in both flavors.

Azure OpenAI's OpenAI-compatible `v1` API (`{endpoint}/openai/v1/...`) returns the same `usage` shape as the direct OpenAI API, so this is a straight gap, not a behavioral difference between the two APIs.

## Why this matters

Without this mapping, Azure OpenAI usage with prompt caching is extracted with `cache_write_tokens` always 0 (dropped, not just unpopulated), so the cache-write premium (which several OpenAI-family models price *higher* than plain input, e.g. gpt-5.6-luna: input $0.20/Mtok vs cache_write $0.25/Mtok) is silently absorbed into the plain input rate instead — under-pricing every Azure request that writes to the prompt cache.

## Verification against live Azure traffic

Confirmed against real Azure OpenAI traffic on both flavors (gpt-5.6-luna, same-prefix cold/warm request pair on each endpoint):

**Responses API** (`{endpoint}/openai/v1/responses`):
```json
// cold (cache write)
{"input_tokens": 8823, "input_tokens_details": {"cache_write_tokens": 8820, "cached_tokens": 0}, ...}
// warm (cache read, same prefix)
{"input_tokens": 8823, "input_tokens_details": {"cache_write_tokens": 17, "cached_tokens": 8803}, ...}
```

**Chat Completions API** (`{endpoint}/openai/v1/chat/completions`):
```json
// cold (cache write)
{"prompt_tokens": 8823, "prompt_tokens_details": {"cache_write_tokens": 8820, "cached_tokens": 0, "audio_tokens": 0}, ...}
// warm (cache read, same prefix)
{"prompt_tokens": 8823, "prompt_tokens_details": {"cache_write_tokens": 17, "cached_tokens": 8803, "audio_tokens": 0}, ...}
```

Both flavors show the same pattern (`cached_tokens + cache_write_tokens ≈ input/prompt_tokens`, identical token counts across both endpoints for the same prefix), confirming `cache_write_tokens` is populated by Azure exactly like it is by direct OpenAI in both API flavors, and that it should be extracted the same way in both.

## Change

Mirrors the existing `openai` provider's mapping exactly — one line added to each of the `azure` provider's `chat` and `responses` extractors in `prices/providers/azure.yml`. Regenerated `data.py` / `data.ts` / `new_data/v2/*` / `tests/dataset/usages.json` via `make build` and `python tests/dataset/extract_usages.py`.

Also added `test_azure_cache_write_tokens` in `tests/test_extract_usage.py`, mirroring the existing `test_openai_cache_write_tokens`, so the mapping is covered by the test suite rather than only having been checked manually.

## Test plan

- [x] `make test` (1172 passed, 42 xfailed, 100% coverage on `packages/python/**`/`tests/**`)
- [x] `make lint`, `make typecheck`
- [x] `npm run ci` (1871 passed, 1 skipped, build/typecheck/lint clean)
- [x] `tests/dataset/usages.json` regenerated and diff reviewed — Azure `chat`/`responses` entries now correctly reflect cache-write pricing instead of the plain input rate
- [x] New `test_azure_cache_write_tokens` (chat + responses) added and passing
- [x] Both flavors verified against live Azure OpenAI traffic (see above)

## AI Disclaimer

This PR was developed with the assistance of Claude. I've reviewed and verified the changes, including validating both extractor mappings against real Azure OpenAI API traffic.


Kludex:
Thanks, this looks right - Azure's v1 API returns the same `usage` shape as OpenAI, and the `openai` provider already maps this field (#463).

I pushed two commits to your branch:
- Merged `main` (the branch was 22 commits behind and the generated `prices/new_data/v2/*.json` conflicted), then re-ran `make build` and `tests/dataset/extract_usages.py`.
- Pointed `test_azure_cache_write_tokens` at `gpt-5.6-sol` instead of `o1`. Azure has no `cache_write_mtok` for `o1`, so the `0.04809` assertion was the same price with or without the cache-write tokens; `gpt-5.6-sol` resolves through Azure's `fallback_model_providers: [openai]` and the `0.015944` total now proves the premium is applied.

PR #674, opened 2026-09-02
Match Pydantic AI provider ID aliases

### LINKED ISSUE #673: Recognize Pydantic AI provider ID aliases
Pydantic AI uses two provider references that differ from the canonical IDs in genai-prices:

- `azure-responses` is the model ID prefix for Azure Responses. The current string-profile path passes it to genai-prices, where provider lookup misses `azure`.
- `gateway/groq` is used by direct callers and older Pydantic AI versions. Current Pydantic AI normalizes it to `groq` before lookup, but genai-prices does not recognize the unnormalized reference itself.

Expected exact matches:

| provider reference | provider |
|---|---|
| `azure-responses` | `azure` |
| `gateway/groq` | `groq` |

Use exact matches only. `huggingface/groq` must continue to resolve to `huggingface_groq`, and unrelated strings such as `gateway/groq-other` must remain unmatched.

No model or pricing data changes are needed.

adtyavrdhn:
Adds two exact provider-reference aliases:

- `azure-responses` resolves to `azure`, fixing the current Pydantic AI string-profile lookup.
- `gateway/groq` resolves to `groq` for direct callers and older Pydantic AI versions. Current Pydantic AI normalizes this reference before lookup.

Both use `equals`, not substring matching. Tests pin the two aliases, preserve `huggingface/groq` as `huggingface_groq`, and reject suffixed near-matches. Generated Python, JavaScript, and Go data is updated by the standard build.

No model or pricing data changes.

Closes #673


adtyavrdhn:
Closing for now. Pydantic AI should normalize its own model and routing identifiers before passing canonical provider IDs to genai-prices. We can revisit this if genai-prices adopts an explicit policy for integration-specific aliases.

PR #676, opened 2026-09-02
Split AWS GPT-5.6 (Luna, Sol, Terra) prices into global and regional

belboo:
## Problem

Bedrock serves the GPT-5.6 models (Luna, Sol, Terra) only through an inference profile, so every request on the `bedrock-runtime` endpoint names a **prefixed** model id — `global.openai.gpt-5.6-luna`, `us.openai.gpt-5.6-luna`, `in.openai.gpt-5.6-luna`. The bare id `openai.gpt-5.6-luna` reaches the model only on the in-Region `bedrock-mantle` endpoint.

The current entries match with `equals: openai.gpt-5.6-<model>`, so none of the prefixed ids resolve and that traffic is costed at zero. We hit this on `global.openai.gpt-5.6-luna` through the Pydantic AI Gateway: real spend, no cost recorded.

The prefixes also carry different prices. **Global cross-Region inference is exactly 10% below in-Region and Geo on every dimension**, for all three models — the same shape as the `global.` / `regional.` split already applied to `claude-sonnet-5` in this file.

## Change

Replace the three `equals:`-matched entries with six:

| removed | added |
|---|---|
| `openai.gpt-5.6-luna` | `global.openai.gpt-5.6-luna`, `regional.openai.gpt-5.6-luna` |
| `openai.gpt-5.6-sol` | `global.openai.gpt-5.6-sol`, `regional.openai.gpt-5.6-sol` |
| `openai.gpt-5.6-terra` | `global.openai.gpt-5.6-terra`, `regional.openai.gpt-5.6-terra` |

The `regional.*` matchers keep the bare id working (`starts_with: openai.gpt-5.6-<model>`) and add both documented Geo prefixes — `us.` for all three, `in.` for Luna and Terra. Those are the only Geo profiles the model cards list; Sol has no India profile. They also accept the bare `gpt-5.6-<model>` form, following the convention from #306 that the `regional.anthropic.*` entries use: when the caller says the provider is Bedrock, the model name is accepted with the vendor prefix stripped. The existing dated price history is carried over unchanged.

## Prices

Per 1M tokens, Standard tier, below 272K input tokens. Every current number is the AWS model card's own figure. The one derived block is the pre-cut global Sol rate (2026-08-17 to 2026-08-21), which AWS never published; it applies the same 10% global discount to the pre-cut in-Region rate. The long-context tier (2x input and cache, 1.5x output) begins at exactly 272K input tokens, so its tier starts are encoded as `271999` — the pricing engines select a tier when the token count is *greater than* `start`. That matches `openai.yml`; the pre-existing `aws.yml` entries used `272000`, which put the boundary one token late, so `test_aws_gpt_5_6_context_boundary` and `test_aws_gpt_5_6_price_change` move with it.

**Luna** — [model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-luna.html)

| | input | cache write (30m) | cache read | output |
|---|---|---|---|---|
| In-Region / Geo | 0.22 | 0.275 | 0.022 | 1.32 |
| Global | 0.20 | 0.25 | 0.02 | 1.20 |

**Sol** — [model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-sol.html)

| | input | cache write (30m) | cache read | output |
|---|---|---|---|---|
| In-Region / Geo | 4.40 | 5.50 | 0.44 | 22.00 |
| Global | 4.00 | 5.00 | 0.40 | 20.00 |

**Terra** — [model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-terra.html)

| | input | cache write (30m) | cache read | output |
|---|---|---|---|---|
| In-Region / Geo | 2.20 | 2.75 | 0.22 | 13.20 |
| Global | 2.00 | 2.50 | 0.20 | 12.00 |

Cross-check: the Global column equals OpenAI's own list price already recorded in [`prices/providers/openai.yml`](prices/providers/openai.yml) (Luna 0.2 / 0.25 / 0.02 / 1.2; Sol 4 / 5 / 0.4 / 20), and the In-Region column is that x1.1.

## Sol price cut, 2026-08-21

`openai.gpt-5.6-sol` currently holds 5.5 / 6.875 / 0.55 / 33 with `prices_checked: 2026-08-10`. Those were correct when written, but AWS cut Sol prices on 2026-08-21 — 20% off input, 33.3% off output — so they are now stale rather than wrong. Per the contributor guidance, this is recorded as a dated conditional entry (`start_date: 2026-08-21`) with the pre-cut rates preserved, not an overwrite.

Ref: [Amazon Bedrock announces reduced pricing for OpenAI GPT-5.6 Sol](https://aws.amazon.com/about-aws/whats-new/2026/08/bedrock-openai-gpt-56-sol-reduced-pricing/)

The same cut applies to the global entry, which therefore also carries two dated blocks: cross-Region inference for the GPT-5.6 models launched on 2026-08-17, four days before the cut, so there is a short window at the pre-cut global rate. Luna and Terra need no such history — their 2026-07-30 price cuts predate global availability.

Ref: [Amazon Bedrock now supports cross-Region inference for OpenAI models](https://aws.amazon.com/about-aws/whats-new/2026/08/amazon-bedrock-cross-region-openai-v2/)

## Sources

- Luna model card: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-luna.html
- Sol model card: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-sol.html
- Terra model card: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-terra.html
- Cross-Region inference launch: https://aws.amazon.com/about-aws/whats-new/2026/08/amazon-bedrock-cross-region-openai-v2/
- Sol price cut: https://aws.amazon.com/about-aws/whats-new/2026/08/bedrock-openai-gpt-56-sol-reduced-pricing/
- Luna and Terra price cut: https://aws.amazon.com/about-aws/whats-new/2026/07/openai-gpt-terra-luna-pricing-bedrock/
- Long-context tier: https://aws.amazon.com/about-aws/whats-new/2026/08/gpt-sol-terra-luna-long-context-bedrock/

Note the AWS Price List API and the pricing-page metered-unit map carry only the GovCloud GPT-5.6 rows, so the model cards are the authoritative commercial-region source. GovCloud rates (1.2x commercial in-Region) are not modelled here, as before.

## Tests

`test_aws_gpt_5_6_context_boundary` gains eight cases covering the `global.` and `us.`/`in.` prefixes and the bare `gpt-5.6-*` form alongside the vendor-qualified ids, so the tier boundary is pinned on every id shape. The Sol expectation moves from 5.5 / 11 to 4.4 / 8.8, reflecting the 2026-08-21 cut.

`make build`, `make test` (100% coverage), `make lint`, `make typecheck`, `go test ./...` and `npm run ci` all pass.

<a href="https://cubic.dev/pr/pydantic/genai-prices/pull/676?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


prices/providers/aws.yml
```diff
@@ -313,6 +313,144 @@ models:
...
+      1.5x. The 80% Luna price cut of 2026-07-30 predates global availability, so there is no earlier
+      global rate to record. Refs:
+      https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-luna.html,
+      https://aws.amazon.com/about-aws/whats-new/2026/08/amazon-bedrock-cross-region-openai-v2/,
+      https://aws.amazon.com/about-aws/whats-new/2026/08/gpt-sol-terra-luna-long-context-bedrock/
+    prices:
+      input_mtok:
+        base: 0.2
+        tiers:
+          - start: 272000
```
alexmojaki:
```suggestion
          - start: 271999
```

and so on for the rest, see openai.yml

prices/providers/aws.yml
```diff
@@ -1155,6 +1167,166 @@ models:
       cache_read_mtok: 0.22
       output_mtok: 11
 
+  - id: regional.openai.gpt-5.6-luna
+    name: GPT-5.6 Luna (regional)
+    match:
+      or:
+        - starts_with: openai.gpt-5.6-luna
+        - contains: us.openai.gpt-5.6-luna
+        - contains: in.openai.gpt-5.6-luna
```
alexmojaki, line 1178: why not include more regions, similar to anthropic models?

should anthropic models also have `in.`?

should this have `starts_with: gpt-5.6-luna` similar to `starts_with: claude-sonnet-5` above?

prices/providers/aws.yml
belboo: Done — all 32 tier starts in the six new entries are now `271999`, and I carried your `openai.yml` wording into their `price_comments` ("tier starts are encoded as 271999 because the pricing engines select a tier when the token count is greater than start").

One consequence worth naming, since it changes behaviour that already shipped: `aws.yml` on `main` used `272000` for these models (12 tier starts), so the boundary moves one token earlier. A request with exactly 272,000 input tokens now bills at long-context rates on AWS, as it already did on the `openai` provider. `test_aws_gpt_5_6_context_boundary` now pins `(271_999 → short, 272_000 → long)` and the `Tier(start=...)` expectations in `test_aws_gpt_5_6_price_change` moved with it.

`openai.yml`'s reading is the one backed by the source — OpenAI lists the standard rates for prompts with **<272K** input tokens — so the old `aws.yml` values were the off-by-one, not this change.

prices/providers/aws.yml
belboo: These three models have only two Geo profiles between them:

| model | Geo inference ID | Global |
|---|---|---|
| [Luna](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-luna.html) | `us.openai.gpt-5.6-luna` in the commercial Regions, `in.openai.gpt-5.6-luna` in the India Regions | `global.openai.gpt-5.6-luna` |
| [Sol](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-sol.html) | `us.openai.gpt-5.6-sol` | `global.openai.gpt-5.6-sol` |
| [Terra](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-terra.html) | `us.openai.gpt-5.6-terra` in the commercial Regions, `in.openai.gpt-5.6-terra` in the India Regions | `global.openai.gpt-5.6-terra` |

No `eu.`, `apac.`, `jp.`, `au.` or `us-gov.` profile exists for any of them — outside the US and India Regions the only cross-Region option on `bedrock-runtime` is Global, and the bare id reaches the model only in-Region on `bedrock-mantle`. So the enumeration in the PR is already complete; that's why it's shorter than the Anthropic ones.

One gap: GovCloud is priced differently apparently, but is not obvious from the model identifier - i.e. indistinguishable from a commercial in-Region request by model ref. Same may be true for `regional.anthropic.*` though, so an open question.

belboo:
Pushed. Rebased onto `main` (picking up GPT-6 Astra #680, GitHub Copilot #683 and Gemini 3.8 Flash #678 — only the generated payloads conflicted, resolved by regenerating with `make build`), then addressed both threads:

- tier starts → `271999`, with `openai.yml`'s explanation carried into `price_comments`; the two AWS GPT-5.6 boundary tests move with it
- `starts_with: gpt-5.6-<model>` added to all three `regional.*` entries
- `us.`/`in.` coverage confirmed complete against the model cards — no other Geo profile exists for these models
- PR body corrected where it said the long-context tier was "encoded as it already was"

Local: `make build` clean, `make test` 1582 passed at 100% coverage, `npm run ci` 2039 passed, `make lint` / `make typecheck` / `make test-go` clean. The only golden-dataset change is 11 bare `gpt-5.6-*` bodies gaining an `aws` price.

I left the second thread open — the "should Anthropic models also have `in.`?" half is a question about entries this PR doesn't touch (answer: no Anthropic card lists an India profile), and the pruning it suggests is yours to call.

prices/providers/aws.yml
belboo: Correction to my reply above: I've dropped the vendor-less `starts_with: gpt-5.6-<model>` matchers again, so this PR now only touches Bedrock-shaped ids.

I had added them to mirror `regional.anthropic.*`, which carries `claude-sonnet-5` beside `anthropic.claude-sonnet-5`. The cost only showed up in `tests/dataset/usages.json`: a bare `gpt-5.6-*` ref is an OpenAI-shaped id, and several recorded bodies in that dataset come from OpenAI cassettes (e.g. `test_openai_responses_prompt_cache_e2e.yaml` with `"model": "gpt-5.6-sol"`). Because the dataset asks every provider's extractor what it would charge, adding the matcher gave `aws` a price for 11 of those bodies — a +22-line diff in a generated cross-language golden file, for ids Bedrock never serves under that name.

`tests/dataset/usages.json` is now byte-identical to `main`, and the PR is back to just the Bedrock GPT-5.6 split. The `271999` tier starts from your other comment stay, as does the boundary test.

Worth noting the same effect already exists for the Anthropic entries — their bare `claude-*` matchers are why `aws` prices bare Anthropic refs in that dataset today. Whether that's intended or worth pruning is a separate question; happy to open a follow-up if you want it changed.

tests/test_price_calc.py
```diff
@@ -496,13 +496,22 @@ def test_aws_gpt_5_6_price_change(model_ref: str, request_timestamp: datetime, e
...
+        ('openai.gpt-5.6-sol', Decimal('4.4'), Decimal('8.8')),
         ('openai.gpt-5.6-terra', Decimal('2.2'), Decimal('4.4')),
         ('openai.gpt-5.6-luna', Decimal('0.22'), Decimal('0.44')),
+        ('global.openai.gpt-5.6-sol', Decimal('4'), Decimal('8')),
+        ('global.openai.gpt-5.6-terra', Decimal('2'), Decimal('4')),
+        ('global.openai.gpt-5.6-luna', Decimal('0.2'), Decimal('0.4')),
+        ('us.openai.gpt-5.6-sol', Decimal('4.4'), Decimal('8.8')),
+        ('us.openai.gpt-5.6-terra', Decimal('2.2'), Decimal('4.4')),
+        ('in.openai.gpt-5.6-luna', Decimal('0.22'), Decimal('0.44')),
+        ('gpt-5.6-sol', Decimal('4.4'), Decimal('8.8')),
```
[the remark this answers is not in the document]
adtyavrdhn: Following Claude's precedence here

PR #690, opened 2026-09-08
fix(aws): match unversioned regional fable model ids

### LINKED ISSUE #689: regional fable 5 ids without a version suffix fail aws price lookup
`genai-prices 0.1.6` and current main fail to price `us.anthropic.claude-fable-5` with `provider_id='aws'`. The AWS regional price entry exists, but its US/EU geographic match clauses require a `-v1` suffix. The unprefixed and global names already resolve.

```python
from genai_prices import Usage, calc_price
calc_price(Usage(input_tokens=1000, output_tokens=100),
           model_ref='us.anthropic.claude-fable-5', provider_id='aws')
# LookupError: Unable to find model ... in aws
```

AWS documents the unversioned US identifier in the [Fable 5 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-fable-5.html). We also have the exact `eu.anthropic.claude-fable-5` identifier in an existing application catalog; that configured alias should resolve to regional pricing without changing inference routing. The current AWS model card does not list the EU identifier, so pricing coverage should not be taken as proof of regional inference availability.

Expected: exact US/EU aliases resolve to the existing regional entry, while global requests retain their distinct rates. Use exact matches to avoid capturing Fable 5.1. No rate changes are needed.

kazmer97:
Add exact matches for unversioned US/EU Fable 5 IDs to the existing AWS regional price entry. The US ID is documented by AWS; the EU alias is used by an existing application catalog. Pricing coverage does not change inference routing or assert regional model availability.

Global and unprefixed lookups retain their existing rates. Exact aliases avoid capturing Fable 5.1. Regenerate all bundled/live data and add Python/JavaScript regressions covering ordinary input, cache reads/writes, and output pricing for all four forms.

Fixes #689.

Rebased onto current upstream main and regenerated the bundled/live data to resolve conflicts, preserving its latest model additions.

Validation: Python suite (1,579 passed, 39 expected failures, 100% coverage), Python Ruff and BasedPyright, JavaScript build/type/lint/coverage suite (2,043 passed, 1 skipped), and Go race/coverage tests (92.9% coverage).


kazmer97:
>codex

Conflicts are resolved at `bfbb884f85e7a4fdd460da5e10590741d1d38a0c`; canonical generated data is rebuilt on current main. Local Python (1,579 passed, 39 expected failures, 100% coverage), JavaScript (2,043 passed), Go, types and pre-commit all pass.

The [hosted Python 3.14 job](https://github.com/pydantic/genai-prices/actions/runs/34252224967/job/102149180616) also passed all 1,579 tests and 100% coverage. Its failure is in `actions/upload-artifact` afterward: `Failed to FinalizeArtifact` returned HTTP 403 from an intermediary. A retry through the API was denied with `Must have admin rights to Repository`. An upstream maintainer needs to retry that job; no test assertion or pricing-data change failed.

